PR #16, opened 2020-10-05
use g4 instead of p3

ArangoGutierrez:

### DIFF
--- a/roles/openshift_node/tasks/create_machine.yml
+++ b/roles/openshift_node/tasks/create_machine.yml
@@ -9,7 +9,7 @@
 
 - name: Check machine is provisioned
   shell: >
-    oc get machines --kubeconfig={{ kubeconfig }} -n openshift-machine-api |grep p3.2xlarge |awk '{ print $2 }'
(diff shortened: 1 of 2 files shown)


roles/openshift_node/tasks/create_machine.yml
```diff
@@ -9,7 +9,7 @@
 
 - name: Check machine is provisioned
   shell: >
-    oc get machines --kubeconfig={{ kubeconfig }} -n openshift-machine-api |grep p3.2xlarge |awk '{ print $2 }'
+    oc get machines --kubeconfig={{ kubeconfig }} -n openshift-machine-api |grep {{ gpu_instance_type }} |awk '{ print $2 }'
```
kpouget, line 12: change `{{ gpu_instance_type` to `'{{ gpu_instance_type }}'` to avoid possible misuse/mistake

kpouget (commented):
* the first patch hunk changes `s/p3.2xlarge/g4dn.xlarge` as per the title,
* the second patch hunk uses the `{{ gpu_instance_type }}` (referring to ^^^) instead of hard-coded value `p3.2xlarge`
  * the reference to the template should be protected to avoid any possible misuse/mistake

sjug:
@ArangoGutierrez Lets combine this with https://github.com/openshift-psap/ci-artifacts/pull/17

PR #19, opened 2020-10-14
Use UBI as base for Dockerfile

ArangoGutierrez:
This patch changes the base image for the Dockerfile to ubi8 to make it
more in-line with OpenShift.

Also removes OC cli install as a Dockerfile step, now that we can
dynamically inject the oc cli as part of multistage tests


jmencak (commented):
I don't feel qualified to approve, however, I've built this image successfully.  Not sure I understand why we don't need the openshift-clients now.  Also, I'd recommend consistent spacing/indentation in the Dockerfile before merging.

build/Dockerfile
```diff
@@ -13,20 +11,17 @@ LABEL 	io.k8s.display-name="OpenShift PSAP Ci artifacts" \
...
+      glibc-langpack-en \
+      git make jq vim wget \
+      python3 python3-devel python3-pip python3-setuptools \
+      && pip3 install --no-cache-dir --upgrade setuptools pip \
+      && pip3 install --no-cache-dir \
+      openshift \
+      ansible==2.9.* \
+      && yum clean all \
+      && rm -rf $HOME/.cache \
+      && rm -rf /var/cache/yum
```
kpouget, line 24: just a style nit, I would cahnge a bit the indentation to better distinguish multiple args (yum/pip3) and multiple commands, something like that:

```suggestion
RUN yum install -y \
        glibc-langpack-en \
        git make jq vim wget \
        python3 python3-devel python3-pip python3-setuptools \
 && pip3 install --no-cache-dir --upgrade setuptools pip \
 && pip3 install --no-cache-dir \
           openshift \
           ansible==2.9.* \
 && yum clean all \
 && rm -rf $HOME/.cache \
 && rm -rf /var/cache/yum
```

ArangoGutierrez:
Test will continue to fail until we can merge https://github.com/openshift/release/pull/12835 
once we can get a `lgtm`here this will fix failing tests for that PR

build/Dockerfile
```diff
@@ -13,21 +11,18 @@ LABEL 	io.k8s.display-name="OpenShift PSAP Ci artifacts" \
...
-    && yum update -y \
-    && yum install --setopt=tsflags=nodocs -y \
-      jq wget git make vim \
-      inotify-tools ansible \
-      python3 openssh-clients \
+RUN yum install -y \
+        glibc-langpack-en \
+        git make jq vim wget \
+        python3 python3-devel python3-pip python3-setuptools \
+ && pip3 install --no-cache-dir --upgrade setuptools pip \
```
sjug, line 18: Can we move `&&` to the end of the previous line?

sjug (changes requested):
Just one thing IMO

build/Dockerfile
ArangoGutierrez: done

playbooks/gpu_machineset_aws_template.json
```diff
@@ -0,0 +1 @@
...ole": "worker", "machine.openshift.io/cluster-api-machine-type": "worker", "machine.openshift.io/cluster-api-machineset": "ed-4-6-dev-gtpbh-worker-us-east-1a"}}, "spec": {"metadata": {}, "providerSpec": {"value": {"ami": {"id": "ami-009cbe327d180d666"}, "apiVersion": "awsproviderconfig.openshift.io/v1beta1", "blockDevices": [{"ebs": {"encrypted": true, "iops": 0, "kmsKey": {"arn": ""}, "volumeSize": 120, "volumeType": "gp2"}}], "credentialsSecret": {"name": "aws-cloud-credentials"}, "deviceIndex": 0, "iamInstanceProfile": {"id": "ed-4-6-dev-gtpbh-worker-profile"}, "instanceType": "m4.large", "kind": "AWSMachineProviderConfig", "metadata": {"creationTimestamp": null}, "placement": {"availabilityZone": "us-east-1a", "region": "us-east-1"}, "securityGroups": [{"filters": [{"name": "tag:Name", "values": ["ed-4-6-dev-gtpbh-worker-sg"]}]}], "subnet": {"filters": [{"name": "tag:Name", "values": ["ed-4-6-dev-gtpbh-private-us-east-1a"]}]}, "tags": [{"name": "kubernetes.io/cluster/ed-4-6-dev-gtpbh", "value": "owned"}], "userDataSecret": {"name": "worker-user-data"}}}}}}, "status": {"availableReplicas": 1, "fullyLabeledReplicas": 1, "observedGeneration": 1, "readyReplicas": 1, "replicas": 1}}
```
jmencak, line 1: Agree with the bot, no EOL is a *bad* thing, same below.

ArangoGutierrez: Ups those files are autogenerated should be there, removed

build/Dockerfile
```diff
@@ -13,20 +11,17 @@ LABEL 	io.k8s.display-name="OpenShift PSAP Ci artifacts" \
...
+        glibc-langpack-en \
+        git make jq vim wget \
+        python3 python3-devel python3-pip python3-setuptools && \
+        pip3 install --no-cache-dir --upgrade setuptools pip && \
+        pip3 install --no-cache-dir \
+        openshift \
+        ansible==2.9.* && \
+        yum clean all && \
+        rm -rf $HOME/.cache && \
+        rm -rf /var/cache/yum
```
kpouget, line 24: I think the indentation got lost since a previously version?

ArangoGutierrez: since I moved the `&&` to the line before, now looks like this, but I get what you mean
fixing

ArangoGutierrez: Done

jmencak (changes requested):
Can we please have some consistency in the Dockerfile?  Sometimes there are tabs, sometimes spaces.  Sometimes we have '&&' at the end of the line, sometimes at the beginning of the line.  Sometimes we indent by 4 spaces, sometimes by one...  These are all "nits", but I'm afraid they hurt (probably not only my) eyes.

ArangoGutierrez:
> Can we please have some consistency in the Dockerfile? Sometimes there are tabs, sometimes spaces. Sometimes we have '&&' at the end of the line, sometimes at the beginning of the line. Sometimes we indent by 4 spaces, sometimes by one... These are all "nits", but I'm afraid they hurt (probably not only my) eyes.

Agree with you Jiri, only me to blame, I think I have diff settings on VIM on my laptop and PC, that creates some indentation misleading here

ArangoGutierrez:
Changed to all tabs

PR #31, opened 2021-01-18
build: do not assume fixed user ID

kpouget:
The directives `ENV USER_UID=1001` and `USER ${USER_UID}` are not
always respected by OpenShift, in particular in Prow CI, where a
random UID is used.


ArangoGutierrez (approved):
/lgtm

PR #35, opened 2021-01-20
roles/openshift_node: fix issue with GPU node scale-up

kpouget:
The code was previously spinning up 2 replicas of GPU machines,
because the source machineset (the default worker one) had 2 replicas
by default.

The task in charge of checking the GPU node deployment was failing
because it was seeing 2x 'Running' while expecting only one.

Also fix 'deff' typo and remove trailing while spaces.


ArangoGutierrez (approved):
Thanks for this @kpouget 
/lgtm

PR #36, opened 2021-01-21
Remove build/root/usr/local/bin/entrypoint

kpouget:
In the Prow CI environment, the container's ENTRYPOINT is never
executed (so the Pod is launched with `spec.containers.command` set),
so no reason to keep this `entrypoint` file here.

Instead, set `INSIDE_CI_IMAGE=y` in the Dockerfile and inside `run`
command, print out all the relevant context information:
- git show
- oc version
- $0 $@ of the `run script`


ArangoGutierrez (approved):
/lgtm

PR #53, opened 2021-02-12
Add GPU-Operator deploy/undeploy from helm

kpouget:
and fix `toolbox/gpu-operator/run_ci_checks.yml` extension

and remove ansible variable `kubeconfig`, most of the commands do not use it


toolbox/gpu-operator/list_version_from_helm.sh
```diff
@@ -0,0 +1,46 @@
+#! /bin/bash
+
+HELM_REPO_NAME="nvidia"
+HELM_REPO_ADDR="https://nvidia.github.io/gpu-operator"
+HELM_REPO_PROJ="gpu-operator"
+
+# https://stackoverflow.com/a/21189044/341106
+function parse_yaml {
```
ArangoGutierrez, line 8: Kevin , now that we are open to using some python here, can't we have an ansible module with python for this? 
I would feel more comfortable parsing YAML with python than with bash. not for this PR, but just a thought for future PR's since you are good with python

ArangoGutierrez: I would this entire bash could be a nice python

ArangoGutierrez (approved):

toolbox/gpu-operator/list_version_from_helm.sh
kpouget: yes indeed,
I'll have a look at Python+Ansible ... for the follow-up PRs ;-)

ArangoGutierrez (approved):
/lgtm

PR #54, opened 2021-02-15
More verbose failure when CR cannot be created

kpouget:


ArangoGutierrez:
Relaunching all tests to see if this PR doesn't need a forced rebase, and if it still works after the last couple merges recently today

kpouget:
wow, things went better than expected for 1.4:

> /usr/local/bin/run: line 121: exec: exit: not found

only `exec exit 0` failed ... because `exit` isn't a command -,-

pushed the fix, so this PR can go in from my POV

ArangoGutierrez (approved):
/lgtm

PR #61, opened 2021-02-18
Create GitHub Action to keep release-* branches synchronized

kpouget:


kpouget:
@dagrayvid if you can have a look, first time I touch GitHub actions

I tested on my repo, it seems to work, and it doesn't work without the `--force`, I'm not sure why but it shouldn't hurt here as we never push to `release-*` directly.

.github/workflows/sync-release-branches.yml
```diff
@@ -0,0 +1,31 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
```
ArangoGutierrez, line 17: Can we use fedora:latest? or UBI latest ;) , you know keeping it Red

ArangoGutierrez (changes requested):
Let's not delete the manual script, keep it as a tool in the tool box, leaving it there doesn't hurts.

.github/workflows/sync-release-branches.yml
kpouget: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer
I don't think we can :(

kpouget:
@ArangoGutierrez I put back the `toolbox/git/push-all.sh` script, and I don't think we can have a RH-based runner ...

.github/workflows/sync-release-branches.yml
ArangoGutierrez: I don't understand why we couldn't based on that documentation, what I am missing?

kpouget: wrong anchor, sorry:

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on

```
Available GitHub-hosted runner types are:

-- | --
Windows Server 2019 | windows-latest or windows-2019
Ubuntu 20.04 | ubuntu-latest (see note) or ubuntu-20.04
Ubuntu 18.04 | ubuntu-latest (see note) or ubuntu-18.04
Ubuntu 16.04 | ubuntu-16.04
macOS Big Sur 11.0 | macos-11.0
macOS Catalina 10.15 | macos-latest or macos-10.15


```

ArangoGutierrez: And couldn't we use 
```yaml
jobs:
  my_job:
    container:
      image: fedora:latest
```
? if not I am ok, is just a personal preference

ArangoGutierrez (approved):
/lgtm

PR #67, opened 2021-02-19
.github/workflows/sync-release-branches.yml: pull with --ff-only and show current branch commit

kpouget:
last try, I don't think it will work but I would like to understand what's the top of each branch


ArangoGutierrez (approved):
/lgtm

PR #72, opened 2021-03-01
toolbox: entitlement: new set of tools

kpouget:
```
toolbox/entitlement/deploy.sh
```
deploys either from a pre-prepared YAML resource file (CI) or from a PEM key

```
toolbox/entitlement/undeploy.sh
```
deletes our MachineConfig objects

```
toolbox/entitlement/test.sh
```
tests if `dnf info kernel-core` completes successfully
```
toolbox/entitlement/wait.sh
```
tests if `dnf info kernel-core` completes successfully within `10*30` seconds
```
toolbox/entitlement/inspect.sh
```
tests if `dnf info kernel-core` completes successfully and gathers extra information about the MachineConfig(Pools), to help debugging entitlement problems

### DIFF
--- a/build/root/usr/local/bin/run
+++ b/build/root/usr/local/bin/run
@@ -15,6 +15,7 @@ ci_banner() {
     echo "===> $0 $@ <=="
 
     git show --quiet || echo "Could not access git history ..."
+    echo
     git show HEAD~ --quiet || true
 
     echo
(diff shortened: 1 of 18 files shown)


kpouget:
the test part already works as expected; tomorrow I'll make sure that the `deploy` and `undeploy` script run well,
and I'll integrate these toolbox into the CI `run` entrypoint

I'll also see how to update the `MachineConfig` [files](https://github.com/openshift-psap/ci-artifacts/pull/72/files#diff-928489e558e0d864a5c93a05c7d6e87e1d47f7759a340f4acee49b6f905721f7), @jmencak mentioned last time that version `2.2.0` of the ignition was too old, it should be hard to update it

kpouget:
/test gpu-operator-e2e

CI was stuck, relaunching it

build/root/usr/local/bin/run
```diff
@@ -69,25 +70,26 @@ prechecks() {
...
-    # for the deployment of the entitlement
-
-    if [[ "${SKIP_ENTITLEMENT:-n}" == y ]]; then
-        echo "Skipping entitlement (SKIP_ENTITLEMENT=${SKIP_ENTITLEMENT})"
+    if toolbox/entitlement/test.sh; then
+        echo "Cluster already entitled, skipping entitlement."
         return
     fi
 
+    if [[ "${INSIDE_CI_IMAGE:-n}" != y ]]; then
```
ArangoGutierrez, line 78: I am a  bit lost, why using `INSIDE_CI_IMAGE` as the way to check this bit

playbooks/entitlement.yml
```diff
@@ -0,0 +1,8 @@
+---
+#
+- name: Deploy and test OpenShift entitlemnen
```
ArangoGutierrez, line 3:
```suggestion
- name: Deploy and test OpenShift entitlement
```

build/root/usr/local/bin/run
kpouget: so the idea of the function
```
entitle() {
```

is that we do nothing if the cluster is already entitled (case 1: inside `toolbox/local-ci/deploy.sh`, case 2 : in the CI when we test `master` and `operatorhub` deployments in a row )
```
    if toolbox/entitlement/test.sh; then
        echo "Cluster already entitled, skipping entitlement."
        return
    fi
```
if `run` is executed outside of the CI image (shouldn't actually happened), there's nothing more we can do
```
    if [[ "${INSIDE_CI_IMAGE:-n}" != y ]]; then
        echo "FATAL: Cluster isn't entitled, cannot continue."
        exit 1
    fi
```
if `run` is executed inside the CI image, our entitlement file should be available
```
    PSAP_ENTITLEMENT="/var/run/psap-entitlement-secret/01-cluster-wide-machineconfigs.yaml"
    if [ ! -f "$PSAP_ENTITLEMENT" ]; then
        echo "FATAL: PSAP entitlement resources not found ($PSAP_ENTITLEMENT)"
        exit 1
    fi

    toolbox/entitlement/deploy.sh --machine-configs $PSAP_ENTITLEMENT
    if ! toolbox/entitlement/wait.sh; then
        echo "FATAL: Failed to properly entitle the cluster, cannot continue."
        exit 1
    fi
}
```

but let me change it a little bit

ArangoGutierrez: I get it, but the `INSIDE_CI_IMAGE` check and then failing is a bit miss leading for an external reader trying to contribute to this repo. 
without much context a person can think it will fail inside a ci image 
maybe 
```bash
 if [[ "${INSIDE_CI_IMAGE:-n}" != y ]]; then
        echo "FATAL: This is intended to be use inside a CI cluster......"
        exit 1
fi
```
or something like that, so the user will realize he/she needs to entitle the cluster to run the playbooks manually

ArangoGutierrez (approved):
/lgtm

PR #78, opened 2021-03-04
Add ability to test and wait for NFD-labelled GPU nodes

kpouget:
This PR improves the detection of GPU nodes in the cluster:

instead of relying on AWS Machines and types to detect the presence of GPU nodes, use NFD to ...
1. test if a GPU node is available (see NVIDIA GPU labels [there](https://github.com/NVIDIA/gpu-operator/blob/master/pkg/controller/clusterpolicy/state_manager.go#L26))
2. if not available, scale-up the cluster with AWS MachineSet
3. wait for the GPU node to show up

this PR implements steps 1 and 3.

Incidentally, the PR also removes the strong dependency on AWS infrastructure, which is now only required when no GPU node is available at the beginning of the execution.

### DIFF
--- a/toolbox/nfd/has_gpu_nodes.sh
+++ b/toolbox/nfd/has_gpu_nodes.sh
@@ -1,0 +1,11 @@
+#! /bin/bash -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${THIS_DIR}/../_common.sh
+
+ANSIBLE_OPTS="${ANSIBLE_OPTS} -e install_nfd_operator_from_hub=no"
+ANSIBLE_OPTS="${ANSIBLE_OPTS} -e install_gpu_operator_from_hub=no"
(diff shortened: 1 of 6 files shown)


toolbox/nfd/has_gpu_nodes.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${THIS_DIR}/../_common.sh
```
ArangoGutierrez, line 4: Why are we sourcing if we are not calling any helper func?

ArangoGutierrez (changes requested):

toolbox/nfd/has_gpu_nodes.sh
kpouget: `_common.sh` is a prolog of the script, 
it sets some ansible variables (inventory, logging) and `cd` to the top directory
https://github.com/openshift-psap/ci-artifacts/blob/master/toolbox/_common.sh

ArangoGutierrez (approved):
/lgtm
/hold

PR #89, opened 2021-03-15
Prepare artifact directory in toolbox/_common.sh

kpouget:
Prepare artifact directory in `toolbox/_common.sh`, and create one directory (with a hour/minute timestamp) per toolbox operation, so that we can better understand what logs belong to which toolbox task.


ArangoGutierrez (approved):
/lgtm

PR #99, opened 2021-03-18
Group config files together

kpouget:
* a58d1ac - config: make yamllint.conf visible

---

* d057c85 - Move ansible.cfg to config/ansible.cfg and update accordingly

---


ArangoGutierrez (approved):
/lgtm

kpouget:
Force merged, the `yamllint` configuration was moved so the test could never pass.
Now we should be able to merge https://github.com/openshift/release/pull/16944

PR #102, opened 2021-03-25
.github/workflows: Add ansible-lint checking

kpouget:


kpouget:
@ArangoGutierrez the PR is ready for review, 
see https://github.com/openshift-psap/ci-artifacts/pull/102/checks?check_run_id=2200981025 for an example of the results when the tests are failing.

https://github.com/openshift-psap/ci-artifacts/pull/100 fixes all the issues

@dagrayvid FYI, the before issue was ... that nothing was linted, you guessed right ;-) Changing the linter version to what I have in my workstation fixed the issue, I guess they added the recursive traversal as a default (`playbooks` vs `playbook/**/*.yml`)
thanks for pointing it out!

ArangoGutierrez:
we can merge this PR since Prow Tide doesn't care about GitHub actions 
if I hit lgtm it will pass and merge
WDYT?

kpouget:
@ArangoGutierrez yes we can merge it

ArangoGutierrez (approved):
/lgtm

PR #109, opened 2021-04-15
Store ansible output as a json logfile

kpouget:
The purpose of this PR is to have another logfile using the JSON format, so that we can easily parse and display the result of the test steps in a website, by fetching the JSON logfile and plotting the relevant information in HTML.


ArangoGutierrez (approved):
/lgtm

kpouget:
@ArangoGutierrez I've updated the PR to generate a valid JSON file (instead of only the JSON list content)

/test gpu-operator-e2e

ArangoGutierrez (approved):
/lgtm

PR #118, opened 2021-04-22
fix entitlement file path

BumbleFeng:
fix entitlement file path, according to https://github.com/openshift-psap/ci-artifacts/blob/master/roles/entitlement_deploy/files/mc_rhsm.yml#L17, https://github.com/openshift-psap/ci-artifacts/blob/master/roles/entitlement_deploy/files/mc_pem.yml#L17 and https://github.com/openshift-psap/ci-artifacts/blob/master/roles/entitlement_deploy/files/mc_pem.yml#L35

Signed-off-by: Feng Huang <fehuang@redhat.com>

### DIFF
--- a/roles/entitlement_test_wait/files/entitlement-tester_entrypoint.yml
+++ b/roles/entitlement_test_wait/files/entitlement-tester_entrypoint.yml
@@ -20,7 +20,7 @@ data:
     echo
 
     echo "# md5sum of entitlement file (debug)"
-    if md5sum /etc/rhsm-host/rhsm.conf /etc/pki/entitlement-host/entitlement{,-key}.pem; then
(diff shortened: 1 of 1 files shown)


kpouget:
Hello @BumbleFeng 
```
/etc/rhsm/rhsm.conf /etc/pki/entitlement/entitlement{,-key}.pem
```
this is, in the [MC files](https://github.com/openshift-psap/ci-artifacts/blob/master/roles/entitlement_deploy/files/mc_pem.yml#L35), is the *host* path, whereas this
```
/etc/rhsm-host/rhsm.conf /etc/pki/entitlement-host/entitlement{,-key}.pem
``` 
is where the host secrets are auto-mounted, so this test is consistent with the deployment we use.

Non cluster-wide deployment + test could be added as an alternative way to deploy the entitlement, but myself I never tried it

Issue #129, opened 2021-04-30
Allow scaling a cluster up and down with N nodes


kpouget:
Currently, `toolbox/cluster/scaleup.sh [instance-type]` allows only adding new `MachineSets` with a given `instance-type`.
For testing the GPU Operator support of scale-up and scale-down, we would need to be able to add new GPU nodes to a cluster, and potentially scale it to 0.

So,

- [x] 1/ `toolbox/cluster/scaleup.sh` should be extended (or with another command) to be able to set the number of desired nodes of a given instance type, eg:

```
toolbox/cluster/scaleup.sh <instance-type> # make sure that machines with <instance-type> are available
toolbox/cluster/scaleup.sh <instance-type> N # make sure that N machines with <instance-type> are available
```

- [ ]  2/ the nightly CI entrypoint should be extended to ensure that this capability works properly in the GPU Operator, with tests like:

```
Deploy the GPU Operator when 0 GPU nodes are available
Scale-up the cluster to 1 GPU node, make sure that the GPU of the nodes gets available
Scale-up the cluster to 2 GPU nodes, make sure that the 2 GPUs get available
Scale-up the cluster to 1 GPU node, make sure that the node disappears
```

kpouget:
PR https://github.com/openshift-psap/ci-artifacts/pull/139 merged, now we need to decide what nightly testing we want to do.

kpouget:
/close

`master` is already testing deployment with `0`  GPU and scale up to 1
`operatorhub'` is testing deployment with `1` GPU.

PR #139, opened 2021-05-04
Change scaleup to set_scale - supported scale other than just 1 node

omertuc:
First step for fixing #129

Rename scaleup role to ensure_machineset which makes sure there's at-least one worker machineset with a given instance type.
If there's no such machineset, it'll create one out of an existing worker machineset (like it does now), but now with 0 replicas instead of 1.

Added new role, called set_scale which receives an instance type and a scale param which:

Ensures that the sum of spec.replicas of all machinesets with the given instance type is equal to the scale param.

If it's not, scales down all of them to 0 except for the first one which will have its spec.replicas set to the scale param.

In any case, it waits for all of machinesets to have:
`status.availableReplicas == spec.replicas`
and
`status.replicas == spec.replicas`

Also added `__pycache__` to `.gitignore`


roles/ensure_machineset/scripts/reconfigure_machineset.sh
```diff
@@ -14,4 +14,4 @@ echo ${json_input} \
     | jq --arg machinesetname "${machinesetname}" '.spec.selector.matchLabels."machine.openshift.io/cluster-api-machineset" = $machinesetname' \
     | jq -c --arg machinesetname "${machinesetname}" '.spec.template.metadata.labels."machine.openshift.io/cluster-api-machineset" = $machinesetname' \
     | jq -c 'del(.status)|del(.metadata.selfLink)|del(.metadata.uid)' \
-    | jq '.spec.replicas = 1'
+    | jq '.spec.replicas = 0'
```
omertuc, line 17: set_scale will later on take care of making sure this value is equal to the requested user value.

reconfigure_machineset's job now is to simply make sure a machineset with that given instance type exists, so 0-replica makes more sense

roles/ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -0,0 +1,46 @@
...
+    oc get machinesets -n openshift-machine-api -o
+    jsonpath='{range .items[?(@.spec.template.metadata.labels.machine\.openshift\.io/cluster-api-machine-role=="worker")]}{.metadata.name}{"\n"}{end}'
+  register: oc_get_machinesets
+  failed_when: not oc_get_machinesets.stdout
+
+- name: Create a temporary file for the original MachineSet
+  ansible.builtin.tempfile:
+    state: file
+    suffix: .machineset.temp.yaml
+  register: machineset_tempfile
```
kpouget, line 13: could you remove this task, and use
```
> {{ artifact_extra_logs_dir }}/source_machineset.json
```
instead of 
```
{{ machineset_tempfile.path }}
```
?

(` {{ artifact_extra_logs_dir }}` used to be an optional variable, but now we're heavily relying on it to store the artifacts required to precisely understand what's going on during the toolbox script execution)

roles/ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -0,0 +1,46 @@
...
+    oc get 'machineset/{{ oc_get_machinesets.stdout_lines[0] }}'
+    -n openshift-machine-api
+    -o json
+    > {{ machineset_tempfile.path }}
+
+- name: Create a temporary file for the new MachineSet
+  ansible.builtin.tempfile:
+    state: file
+    suffix: .machineset.temp.yaml
+  register: new_machineset_tempfile
```
kpouget, line 26: same here, 

> {{ artifact_extra_logs_dir }}/new_machineset.json

roles/ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -0,0 +1,46 @@
...
+    state: file
+    suffix: .machineset.temp.yaml
+  register: machineset_tempfile
+
+- name: Store the MachineSet json definition in a temp file
+  shell:
+    oc get 'machineset/{{ oc_get_machinesets.stdout_lines[0] }}'
+    -n openshift-machine-api
+    -o json
+    > {{ machineset_tempfile.path }}
```
kpouget, line 20: can you indent these lines with 3 spaces, to align with `get`

roles/set_scale/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+dependencies:
+  - role: check_deps
+  - role: ensure_machineset
```
kpouget, line 4: hum, I never linked roles this way, I included them explicitly in the `main.yml`
could you add a comment at the beginning of `roles/set_scale/tasks/main.yml ` to mention that this role will be executed before `roles/set_scale/tasks/main.yml ` starts?

toolbox/cluster/set_scale.sh
```diff
@@ -0,0 +1,23 @@
...
+    INSTANCE_TYPE=${1}
+    ANSIBLE_OPTS="${ANSIBLE_OPTS} -e machineset_instance_type=${1}"
+fi
+
+shift
+
+if [ "${1-}" ]; then
+    echo "Setting cluster ${INSTANCE_TYPE} machinesets to have scale '${1}'"
+    ANSIBLE_OPTS="${ANSIBLE_OPTS} -e scale=${1}"
+fi
```
kpouget, line 21: Could you add a ` roles/set_scale/default/config.yml` file to define `machineset_instance_type` and `scale`, with a comment line explaining their purpose, and setting a default value if relevant?

and at the beginning of `roles/set_scale/tasks/main.yml`, make sure that the variables are defined / populated, eg:
```
fail: ...
when: not scale
```

it's not clear to me what you are testing here with the 3 `if`, and which parameters are mandatory, which are optional.

for the mandatory one in the `USAGE`, I use `<name>` and `[name]` for optional ones

build/root/var/lib/ci-artifacts/entrypoints.d/gpu_operator_run_libs.sh
```diff
@@ -13,7 +13,7 @@ prepare_cluster_for_gpu_operator() {
     fi
 
     if ! toolbox/nfd/has_gpu_nodes.sh; then
-        toolbox/cluster/scaleup.sh
+        toolbox/cluster/set_scale.sh g4dn.xlarge 1
```
kpouget, line 16: for testing & coverage purposes, I would prefer this line to be written as
```
toolbox/cluster/ensure_machineset.sh g4dn.xlarge
toolbox/cluster/set_scale.sh g4dn.xlarge 1
```
I saw that `set_scale.sh` calls `ensure_machineset` role as a dependency,
but by calling it explicitly, we ensure that `ensure_machineset.sh` script & playbook don't break without us noticing it!

omertuc: I got rid of `ensure_machineset.sh` because I felt it was a bit of an implementation detail not relevant from a toolbox user POV, so there's no longer anything that can break without us noticing.

Let me know if you disagree with that choice

toolbox/cluster/set_scale.sh
omertuc: Everything is required. The last 2 `if`s are a silly mistake, got rid of them.

build/root/var/lib/ci-artifacts/entrypoints.d/gpu_operator_run_libs.sh
kpouget: indeed, it makes sense not to have `ensure_machineset.sh`

toolbox/cluster/set_scale.sh
kpouget: `${2:-}` is the same as `${2-}` ? .... I need to take time to learn bash extended variable support :#

omertuc: Not sure, just copied it from https://github.com/openshift-psap/ci-artifacts/blob/7d033e63dc650b37d7c634080e9511af4f3d0d1e/toolbox/cluster/scaleup.sh#L6

omertuc:
\test gpu-operator-e2e

roles/ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -0,0 +1,34 @@
...
+    oc get 'machineset/{{ oc_get_machinesets.stdout_lines[0] }}'
+       -n openshift-machine-api
+       -o json
+       > {{ artifact_extra_logs_dir }}/source_machineset.json
+
+- name: Update the MachineSet with the requested instance type ({{ machineset_instance_type }}) and store in a file
+  shell:
+    "{{ reconfigure_machineset }}
+      {{ artifact_extra_logs_dir }}/source_machineset.json
+      {{ machineset_instance_type }}
```
kpouget, line 20: could you add 
```
# The instance type to user for the machineset being scaled
machineset_instance_type: ""
```
to `roles/ensure_machineset/defaults/main/config.yml`, so that the role defines its configuration variables?

playbooks/set_scale.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Ensure cluster machinesets with given instance type have given scale (total number of replicas equals requested number of replicas)
+  hosts: localhost
+  connection: local
+  gather_facts: true
+  roles:
+  - set_scale
```
kpouget, line 7: can you rename the playbook & role to `cluster_set_scale`?

playbooks/ensure_machineset.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Ensures cluster has at least one machineset with the given instance type
+  hosts: localhost
+  connection: local
+  gather_facts: true
+  roles:
+  - ensure_machineset
```
kpouget, line 7: can you rename the playbook and role to `cluster_ensure_machineset`?

playbooks/ensure_machineset.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Ensures cluster has at least one machineset with the given instance type
```
kpouget, line 2: `Ensures that the cluster has ...`?

playbooks/set_scale.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Ensure cluster machinesets with given instance type have given scale (total number of replicas equals requested number of replicas)
```
kpouget, line 2: `Ensures that the cluster has the request number of machines of a given machine-type` ?

roles/set_scale/defaults/main/config.yml
```diff
@@ -0,0 +1,9 @@
+---
+# The instance type to user for the machineset being scaled
+machineset_instance_type: "g4dn.xlarge"
```
kpouget, line 3: I think you can remove the default value here, as the instance type is mandatory in the toolbox script

roles/set_scale/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
+# See meta/main.yml for a list of dependency roles that will be executed prior to this role
+---
+- name: Ensure parameters have been set
+  fail: msg="Bailing out. This role requires 'machineset_instance_type' and 'scale' to be defined"
+  when: not machineset_instance_type or not scale
```
kpouget, line 5: I wonder what happens when `scale = 0`, did you try?

roles/set_scale/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+
+- when: current_replicas_sum != scale
+  name: Change all {{ machineset_instance_type }} machinesets replicas to have sum {{ scale }}
+  block:
+  - name: Get the names of existing worker machinesets with type {{ machineset_instance_type }}
+    command:
+      oc get machineset -n openshift-machine-api -o=jsonpath='{range .items[?(@.spec.template.spec.providerSpec.value.instanceType=="{{ machineset_instance_type }}")]}{.metadata.name}{"\n"}{end}'
+    register: oc_get_machinesets
+    failed_when: not oc_get_machinesets.stdout
+  - name: Set the replicas of all {{ machineset_instance_type }} machinesets to 0 except the first
```
kpouget, line 29: can you add new lines between tasks?

roles/set_scale/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+    command: |
+      oc patch machineset -n openshift-machine-api {{ item }} --patch '{"spec": {"replicas": 0 }}' --type merge
+    loop: "{{ oc_get_machinesets.stdout_lines[1:] }}"
+  - name: Get first machineset
+    set_fact:
+      first_machineset: "{{ oc_get_machinesets.stdout_lines[0] }}"
+  - name: Set the replicas of the first {{ machineset_instance_type }} machineset ({{ first_machineset }}) to {{ scale }}
+    command: >
+      oc patch machineset -n openshift-machine-api {{ first_machineset }}
+      --patch '{"spec": {"replicas": {{ scale }} }}' --type merge
```
kpouget, line 39: I'm not sure modifying *all* the machinesets [with the given machine-type] is the best strategy,
maybe you could change it to something like:

is there only one MachineSet generating type `machine-type`? yes, modify it to `scale`
no --> cannot perform task ==> allow receiving a machineset name change change only this replica to `scale`

roles/set_scale/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+      | jq 'select(.status.availableReplicas != .spec.replicas or .status.replicas != .spec.replicas)' \
+      | jq '{
+         "name": .metadata.name,
+          "expected_replicas": .spec.replicas,
+          "current_replicas": .status.replicas,
+          "current_available_replicas": .status.availableReplicas
+      }'
+    register: non_ready_replicas
+    until: not non_ready_replicas.stdout_lines
+    retries: 120 # TODO: Does this correlate with {{ scale }}? Probably not, otherwise might need to factor it in
```
kpouget, line 60: no, not directly, the machine go live in parallel!

kpouget (changes requested):

playbooks/set_scale.yml
omertuc: I prefer to use [instance type](https://aws.amazon.com/ec2/instance-types/) rather than machine type, it's also the name of the variable (`machineset_instance_type`)

roles/set_scale/tasks/main.yml
omertuc: Good point. Just tested - even when scale is `0`, seems like it's actually being treated as `"0"` which is a non empty string and thus `not scale` is still negative.

roles/set_scale/tasks/main.yml
omertuc: That's what I thought, will remove the TODO because your confirmation made me more confident

roles/set_scale/tasks/main.yml
omertuc: I thought the idea was to make sure the number of machines of given instance type, across the entire cluster, is exactly `scale`.

If we ask the user to provide us with a machineset name that needs to be modified - then the role doesn't really do its job, it only modifies that specific machineset, and doesn't achieve the goal of setting the cluster-wide total to exactly `scale`.

Here's an alternative suggestion - if there are other machinesets in the cluster with this instance type that need to be downscaled to 0 in order to achieve the required cluster-wide `scale`, then the role will ask the user to provide a --force flag telling them it's going to happen, so we don't perform anything extreme without explicit confirmation from the user.

playbooks/set_scale.yml
kpouget: sure, `instance-type` is the correct one, machine-type was a typo :)
my suggestion was more for having a shorter description

roles/set_scale/tasks/main.yml
kpouget: > I thought the idea was to make sure the number of machines of given instance type, across the entire cluster, is exactly `scale`.

The main (and only) requirement is to allow the CI to scale up and down the # of GPU nodes,
so in this environment, we know that only one MachineSet of the given instance-type will be available.
The question could be solved by aborting the play if multiple MachineSets are found.

But I like the `--force` option, it makes sense to me: default behavior is to do things as simple as possible, force behavior makes sure we have the right count, possibility by scaling down some machinesets: :+1:

toolbox/cluster/set_scale.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/bash -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${THIS_DIR}/../_common.sh
+
+if [ -z "${2-}" ]; then
+    echo "USAGE: ${0} <instance_type> <scale>"
+    exit 1
```
kpouget, line 8: Could you add more information about the usage?
in particular, could you mention here some the instance-types that offer GPUs?

and I was wondering how much of it is AWS specific? I was wondering if the name should be `set_scale_aws.sh` or not

roles/cluster_set_scale/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    shell: >
+      oc get machinesets -n openshift-machine-api \
+        -o=jsonpath='{.items[?(@.spec.template.spec.providerSpec.value.instanceType=="{{ machineset_instance_type }}")]}' \
+      | jq 'if .status | has("availableReplicas") then . else . * {"status": {"availableReplicas": 0 }} end' \
+      | jq 'select(.status.availableReplicas != .spec.replicas or .status.replicas != .spec.replicas)' \
+      | jq '{
+         "name": .metadata.name,
+          "expected_replicas": .spec.replicas,
+          "current_replicas": .status.replicas,
+          "current_available_replicas": .status.availableReplicas
```
kpouget, line 59: there is a mis-alignment here, `"name"` is one space before the rest :)

roles/cluster_set_scale/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    set_fact:
+      first_machineset: "{{ oc_get_machinesets.stdout_lines[0] }}"
+
+  - name: Set the replicas of the first {{ machineset_instance_type }} machineset ({{ first_machineset }}) to {{ scale }}
+    command: >
+      oc patch machineset -n openshift-machine-api {{ first_machineset }}
+      --patch '{"spec": {"replicas": {{ scale }} }}' --type merge
+
+- block:
+  - name: Wait for all machinesets with type {{ machineset_instance_type }} to be ready
```
kpouget, line 45: nice, this wait for the Node to be running, I wasn't sure!
Could you add a comment saying that the `.status.availableReplicas` counter tracks the # of nodes running, and not only the # of `Machines`? (I though it would be the latest only)

toolbox/cluster/set_scale.sh
omertuc: The `.sh` file itself can forever be agnostic I believe, all platforms just need an instance type name and a number to scale.

The playbooks however might already be aws-specific, not sure. Since we don't have backwards compatibility concerns, I believe it will be easy to generalize them once we want to, so no need to worry about it for now.

toolbox/cluster/set_scale.sh
```diff
@@ -0,0 +1,45 @@
+#! /bin/bash -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+if [ -z "${2-}" ] && [ -z "${3-}" ]; then
+    echo "USAGE: ${0} <instance_type> <scale> [--force]"
+    echo "
+For a list of AWS instance_types see https://aws.amazon.com/ec2/instance-types/ - check the 'Accelerated Computing' section for instances with GPU
```
kpouget, line 8: can you mention explicitly `g4dn.xlarge` as an example of cheap GPU node, and `m4.large` as the default worker node type (correct me if I'm wrong on this one)

toolbox/cluster/set_scale.sh
```diff
@@ -0,0 +1,45 @@
+#! /bin/bash -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+if [ -z "${2-}" ] && [ -z "${3-}" ]; then
```
kpouget, line 5: I don't think this is the intended tests: if the 2nd and 3rd arguments are empty --> usage

maybe: `[[ -z "${1:-}" || -z "${2:-} || "${3:---force}" != "--force""`?
if 1 or 2 are empty, or 3 isn't set or is equal to `--force`

toolbox/cluster/set_scale.sh
```diff
@@ -0,0 +1,45 @@
...
+set to true. In that case, the number of replicas of the other machinesets will be zeroed before setting the replicas
+of the first machineset to the value of this parameter."
+
+    exit 1
+fi
+
+source ${THIS_DIR}/../_common.sh
+
+INSTANCE_TYPE=${1}
+ANSIBLE_OPTS="${ANSIBLE_OPTS} -e machineset_instance_type=${1}"
```
kpouget, line 28: you can reuse `$INSTANCE_TYPE` here, instead of `${1}`

README.md
```diff
@@ -157,13 +157,12 @@ toolbox/nfd/wait_gpu_nodes.sh
...
 ```
-./toolbox/cluster/scaleup.sh
-```
-   - [x] Specify a machine type in the command-line, and skip scale-up if a node with the given machine-type is already present
-```
-./toolbox/cluster/scaleup.sh <machine-type>
+./toolbox/cluster/set_scale.sh <machine-type> <replicas>
+
+Example usage:
+./toolbox/cluster/set_scale.sh g4dn.xlarge 2
```
kpouget, line 165: maybe you can add `[--force]` here, and a line of comment about it

we'll move that to the documentation soon anyway

kpouget (changes requested):
thanks for the update @omertuc , I pointed out a few nits (mostly in the toolbox script)

toolbox/cluster/set_scale.sh
omertuc: If unspecified, [the installer chooses between m4.large and m5.large according to availability](https://github.com/openshift/installer/blob/7a6374014536d5993e444e961812dd8adb2ed3fd/pkg/asset/machines/worker.go#L159)

kpouget:
Looks good @omertuc, thanks :)

/approve

needs to rebase before LGTM

kpouget:
Looks good @omertuc, thanks :)

/lgtm but it won't merge before another round of rebase + review!

kpouget (approved):
/lgtm

thanks @omertuc for this first PR, that's clean work, really appreciated!

PR #143, opened 2021-05-05
Moving libraries and entitle.sh into /usr/local/bin

dagrayvid:


build/root/usr/local/bin/entitle.sh
```diff
@@ -12,9 +16,6 @@ extract_entitlement_key() {
         > "$key"
 }
 
-THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-cd ${THIS_DIR}/../..
-
```
kpouget, line 17: the file cannot work without something like this:
```
toolbox/entitlement/test.sh
```

this assumes that we are in the `ci-artifacts` main directory
... but actually, maybe we are, as it's also `$HOME` ...

Could you add something like this:
```

if ! [ -f toolbox/entitlement/test.sh ]; then
  echo "FATAL: entitlement scripts not found in $PWD/toolbox/entitlement/"
  echo "INFO: $0 is intended only for running in the 'OpenShift PSAP CI artifacts' image. (INSIDE_CI_IMAGE=$INSIDE_CI_IMAGE)"
  exit 1
fi

```

build/root/usr/local/bin/run
```diff
@@ -103,25 +103,23 @@ fi
 
 set -x
 
-ENTRY_LIB_DIR="/var/lib/ci-artifacts/entrypoints.d"
 case ${target} in
     "gpu-operator")
-        $ENTRY_LIB_DIR/gpu_operator_run_libs.sh "$action" "$@"
```
kpouget, line 109: what do you think about these names?
```
ci_entrypoint_gpu-operator.sh
ci_entrypoint_sro.sh
ci_entrypoint_cluster.sh
```

kpouget (changes requested):
thanks for the PR David, just a few nits to fix and that should be good

I prefer this organization with `entitle.sh` inside the `build/root` instead of the toolbox:)

kpouget:
/lgtm
/approve

thanks for updating it :)

PR #158, opened 2021-05-17
gpu_operator_set_repo-config: fix RHEL 8.4-beta repo-list

kpouget:
I faced 3 issues to get RHEL 8.4 beta repo-list to work as expected in
the GPU Operator driver-container:

a) the driver-container script tries to disable some repo, so if the
file is readonly, it fails (https://gitlab.com/nvidia/container-images/driver/-/issues/25)

b) it needs packages not available in the beta mirror, so these extra
repo must be available, eg (from install of `kernel-devel`):

```
perl-Encode              x86_64  4:2.97-3.el8                            rhel-8-beta-baseos-rpms           1.5 M
perl-Digest-MD5          x86_64  2.55-396.el8                            rhel-8-for-x86_64-appstream-rpms   37 k
openssl                  x86_64  1:1.1.1g-15.el8_3                       rhel-8-for-x86_64-baseos-rpms     707 k
```

c) the default repo cannot be fetched AFAIU, so they need to be masked
--> mount the file in `/etc/yum.repos.d` and force `rhelver=8.3` in the repo URL

### DIFF
--- a/roles/gpu_operator_set_repo-config/tasks/main.yml
+++ b/roles/gpu_operator_set_repo-config/tasks/main.yml
@@ -33,7 +33,6 @@
         [rhel-8-beta-baseos-rpms]
         name = Red Hat Enterprise Linux 8 Beta BaseOS (RPMs)
         baseurl = https://mirror.openshift.com/enterprise/reposync/ci-deps/rhel-8-beta-baseos-rpms/
-        failovermethod = priority
         gpgcheck = 0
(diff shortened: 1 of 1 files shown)


kpouget:
/approve

@dagrayvid this PR modify only the repo-list of RHEL 8.4-beta (OCP 4.8) which is currently broken, so no need to test it before merging

... and hopefully it will fix it :)

dagrayvid:
Makes sense
/lgtm

PR #172, opened 2021-05-21
WIP: Build GPU Operator operands to internal registry and deploy GPU operator with them

omertuc:
WIP - Still need to add a CI job to openshift/release so this can be tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 __pycache__
+/playbooks/cache/
--- a/playbooks/build-operand.yml
+++ b/playbooks/build-operand.yml
@@ -1,0 +1,7 @@
+---
+- name: Build an operand image on the local cluster
+  hosts: localhost
+  connection: local
+  gather_facts: true
+  roles:
+    - role: build-operand
--- a/roles/build-operand/tasks/main.yml
+++ b/roles/build-operand/tasks/main.yml
@@ -1,0 +1,49 @@
+---
+- name: Make sure imagestream name is set
+  when: not build_operand_imagestream_name
+  fail: msg="Failed because build_operand_imagestream_name has to be set"
+
+- name: Make sure build_operand_tag is set
+  when: not build_operand_tag
+  fail: msg="Failed because build_operand_tag has to be set"
+
+- name: Create the namespace manifest
+  template:
+    src: "{{ build_operand_namespace_deployment }}"
+    dest: "{{ artifact_extra_logs_dir }}/{{ build_operand_imagestream_name }}_namespace.yml"
+
+- name: Apply the namespace manifest
+  command:
+    oc apply -f "{{ artifact_extra_logs_dir }}/{{ build_operand_imagestream_name }}_namespace.yml"
+
+- name: Create the imagestream manifest
+  template:
+    src: "{{ build_operand_imagestream }}"
+    dest: "{{ artifact_extra_logs_dir }}/{{ build_operand_imagestream_name }}_imagestream.yml"
+
+- name: Apply the namespace manifest
+  command:
+    oc apply -f "{{ artifact_extra_logs_dir }}/{{ build_operand_imagestream_name }}_imagestream.yml"
+
+- name: Search if the CI image exists
+  command: oc get imagestreamtag \
+              -n "{{ build_operand_namespace }}" \
+              "{{ build_operand_imagestream_name }}:{{ build_operand_tag }}" \
+              -oname --ignore-not-found
+  register: has_ci_image
+
+- name: Build the CI image
+  when: has_ci_image.stdout == ""
+  block:
+  - name: Delete any old image builder manifest
+    command: oc delete -n "{{ build_operand_namespace }}" buildconfig/{{ build_operand_imagestream_name }} --ignore-not-found
+
+  - name: Apply the image builder manifest
+    command:
+      oc apply -f "{{ operand_buildconfig }}"
+
+  - name: Wait for the image to be built
+    shell: oc logs -f "buildconfig/{{ build_operand_imagestream_name }}" -n "{{ build_operand_namespace }}" > /dev/null
+
+- name: Ensure that the image exists
+  command: oc get imagestreamtag -n "{{ build_operand_namespace }}" "{{ build_operand_imagestream_name }}:{{ build_operand_tag }}" -oname
--- a/roles/clean-operand/tasks/main.yml
+++ b/roles/clean-operand/tasks/main.yml
@@ -1,0 +1,10 @@
+---
+- name: Check that build_operand_imagestream_name is defined
+  fail: msg="Bailing out. This play requires build_operand_imagestream_name"
+  when: build_operand_imagestream_name | default('', true) | length == 0
+
+- name: Delete BuildConfig
+  command: oc delete -n "{{ build_operand_namespace }}" buildconfig/{{ build_operand_imagestream_name }} --ignore-not-found
+
+- name: Delete ImageStream
+  command: oc delete -n "{{ build_operand_namespace }}" imagestream/{{ build_operand_imagestream_name }} --ignore-not-found
--- a/roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_clusterpolicy.yml
+++ b/roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_clusterpolicy.yml
@@ -45,8 +45,19 @@
     | jq -r '.metadata.annotations."alm-examples"'
     | jq .[0] > "{{ artifact_extra_logs_dir }}/gpu_operator_clusterpolicy.json"
 
+- name: Optionally call the patch_clusterpolicy_role hook to modify the clusterpolicy
+  include_role:
+    name: "{{ patch_clusterpolicy_role }}"
+  when: patch_clusterpolicy_role | default('', true) | trim != ''
+
+- name: No patch_clusterpolicy_role hook, simply copy the original clusterpolicy
+  command: cp
+    "{{ artifact_extra_logs_dir }}"/gpu_operator_clusterpolicy.json
+    "{{ artifact_extra_logs_dir }}"/gpu_operator_clusterpolicy_patched.json
+  when: not patch_clusterpolicy_role
+
 - name: Create the clusterPolicy CR for the GPU Operator
-  command: oc apply -f "{{ artifact_extra_logs_dir }}/gpu_operator_clusterpolicy.json"
+  command: oc apply -f "{{ artifact_extra_logs_dir }}/gpu_operator_clusterpolicy_patched.json"
 
 - name: Get the GPU Operation version
   include_role:
--- a/build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
+++ b/build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
@@ -4,6 +4,12 @@ set -o pipefail
 set -o errexit
 set -o nounset
 
+build_all_operands() {
+    trap collect_must_gather ERR
+
+    toolbox/gpu-operator/operand-build/build_all.sh
+}
+
 prepare_cluster_for_gpu_operator() {
     ./run_toolbox.py cluster capture_environment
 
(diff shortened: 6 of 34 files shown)


toolbox/gpu-operator/operand-build/build_all.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash -e
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${THIS_DIR}/../_common.sh
+
+operands_dir=${THIS_DIR}/operands
```
omertuc, line 6: We might want to generate the `*.json` files in this directory in a more "smart" way - I stole all the docker build args from the Makefiles of the various operand repos or guessed them by common sense.

toolbox/gpu-operator/operand-build/operands/dcgm_exporter.json
```diff
@@ -0,0 +1,23 @@
+{
+  "build_operand_git_repo": "https://gitlab.com/nvidia/container-toolkit/gpu-monitoring-tools",
```
omertuc, line 2: We might want to fork all the operand repos, to be more in a control.

For example, all the changes we might want to make to support different architectures seamlessly - it might be annoying to wait changes like that to be merged in nvidia upstream.

omertuc: Also, when they're forked - they're less of a moving target, we're more in control of which versions we are testing

toolbox/gpu-operator/operand-build/operands/device_plugin.json
```diff
@@ -0,0 +1,19 @@
...
+  "build_operand_git_ref": "master",
+  "build_operand_namespace": "gpu-operator-resources",
+  "build_operand_imagestream_name": "k8s-device-plugin",
+  "build_operand_tag": "latest",
+  "build_operand_dockerfile": "docker/Dockerfile.ubi8",
+  "build_operand_docker_context": ".",
+  "build_operand_docker_build_args": [
+    {
+      "name": "GOLANG_VERSION",
+      "value": "1.15.8"
```
omertuc, line 12: An example of a "bad" hardcoded value we might want to retrieve in a smarter way, as mentioned in [this comment](https://github.com/openshift-psap/ci-artifacts/pull/172/files#r637164014)

toolbox/gpu-operator/operand-build/operands/toolkit.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+      "name": "NVIDIA_CONTAINER_RUNTIME_VERSION",
+      "value": "3.5.0"
+    }
+  ],
+  "build_operand_extra_docker_strategy": {
+    "from": {
+      "kind": "DockerImage",
+      "name": "nvcr.io/nvidia/cuda:11.2.1-base-ubi8"
```
omertuc, line 34: This avoid [this](https://gitlab.com/nvidia/container-toolkit/container-config/blob/a79c4dcf0c2e2d72f7b3578bc95f206fb05a0ddb/docker/Dockerfile.ubi8#L36) Docker hub pull - but we can only affect the last FROM in a multi-stage build. The [first one](https://gitlab.com/nvidia/container-toolkit/container-config/blob/a79c4dcf0c2e2d72f7b3578bc95f206fb05a0ddb/docker/Dockerfile.ubi8#L15) will still happen from Docker Hub, and we might hit our rate limit if we don't fix that.

Maybe we should ask nvidia to stop using Docker hub upstream, or [fork the repos and fix that ourselves](https://github.com/openshift-psap/ci-artifacts/pull/172/files#r637164737)

roles/build-operand/files/image_builder.yml
```diff
@@ -0,0 +1,28 @@
...
+      ref: "{{ build_operand_git_ref }}"
+    contextDir: "{{ build_operand_docker_context }}"
+  triggers:
+  - type: "ConfigChange"
+  strategy:
+    type: Docker
+    dockerStrategy:
+      buildArgs: {{ build_operand_docker_build_args }}
+      dockerfilePath: "{{ build_operand_dockerfile }}"
+      {{ build_operand_extra_docker_strategy }}
```
omertuc, line 28: This hacky yaml-sed-merge thingy is not appreciated by both the YAML and ansible lints (ansible-lint is running on this file, for some reason...). Would appreciate any tips on how to do this better

toolbox/gpu-operator/operand-build/build_all.sh
```diff
@@ -0,0 +1,12 @@
...
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${THIS_DIR}/../../_common.sh
+
+operands_dir=${THIS_DIR}/operands
+
+for operand_json in $(ls ${operands_dir}); do
+  exec ansible-playbook -e "$(< ${operands_dir}/${operand_json})" playbooks/build-operand.yml &
+done
+
+wait
```
kpouget, line 12: I think you'll need a way to catch errors in `ansible-playbook`

When you run this:
```
set -e
sleep 10 &
sleep 15 &
false &

wait
echo done
```
the error on `false` (= a playbook failing) is swallowed and the script completes successfully

toolbox/gpu-operator/operand-build/clean_all.sh
```diff
@@ -0,0 +1,12 @@
...
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${THIS_DIR}/../../_common.sh
+
+operands_dir=${THIS_DIR}/operands
+
+for operand_json in $(ls ${operands_dir}); do
+  exec ansible-playbook -e "$(< ${operands_dir}/${operand_json})" playbooks/clean-operand.yml &
+done
+
+wait
```
kpouget, line 12: same comments here for the parallel execution of the playbooks and the `wait`

toolbox/gpu-operator/operand-build/operands/dcgm_exporter.json
kpouget: we don't *want* to fork them, but we may *have* to do so :)
it the image cannot be easily built for an architecture and the dockerfile file needs to be fixed, fork the repo and fix what ever has to be fixed; (make sure we agree on the fix) and share the PR with NVIDIA, so that we can get rid of our fork

> Also, when they're forked - they're less of a moving target, we're more in control of which versions we are testing

one thing to take into account somehow it that a given version of the GPU Operator works with a given version of the operands. There might be backward compatibility, but not necessarily forward.

So we need some flexibility in the configuration of the `build_operand_git_ref`, and maybe in the `build_operand_tag`.

toolbox/gpu-operator/operand-build/operands/toolkit.json
kpouget: > Maybe we should ask nvidia to stop using Docker hub upstream, or fork the repos and fix that ourselves

yes, this needs to be discssed with NVIDIA, pulling from DockerHub gets me build failure on Quay.io as well ...

kpouget (commented):
thanks for sharing the WIP @omertuc,
I made some comments but I couldn't look at the ansible role yet

.gitignore
```diff
@@ -1 +1,2 @@
 __pycache__
+/playbooks/cache/
```
kpouget, line 2: I'm wondering how this directory got generated, I don't have it in my system

playbooks/build-operand.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Build an operand image on the local cluster
+  hosts: localhost
+  connection: local
+  gather_facts: true
+  roles:
+    - role: build-operand
```
kpouget, line 7: Could you rename the roles&playbooks to `gpu_operator_$NAME`?

roles/build-operand/files/image_builder.yml
kpouget: can you use the `template` task instead of the `sed` hack?
from my 1st try in `gpu_operator_deploy_from_operatorhub` it works well

however the "template" file shouldn't change much, we'll have to study why it's complaining
it the linter is too strict for what we want, we can consider disabling some rules, it it's enforcing meaningless constraints or if we can't solve the warnings in a clean way

roles/build-operand/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Apply the imagestream manifest
+  shell:
+    set -o pipefail;
+    cat "{{ build_operand_imagestream }}" \
+    | sed 's|{{ '{{' }} build_operand_namespace {{ '}}' }}|{{ build_operand_namespace }}|' \
+    | sed 's|{{ '{{' }} build_operand_imagestream_name {{ '}}' }}|{{ build_operand_imagestream_name }}|' \
+    | oc apply -f-
+
+- name: Search if the CI image exists
+  command: oc get imagestreamtag -n "{{ build_operand_namespace }}" "{{ build_operand_imagestream_name }}:{{ build_operand_tag }}" -oname --ignore-not-found
```
kpouget, line 22: can you split in multiple lines, grouping together flags&values, eg
```
oc get imagestreamtag
   -n "{{ build_operand_namespace }}" 
   "{{ build_operand_imagestream_name }}:{{ build_operand_tag }}"
  -oname 
   --ignore-not-found
```
and I usually use `--ignore-not-found=true` to be more explicit, because all the `oc` commands don't have the same default value. Maybe it's redundant, idk ...

roles/clean-operand/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
```
kpouget, line 1: can you check that `build_operand_imagestream_name` isn't empty?

roles/clean-operand/vars/main/resources.yml
```diff
@@ -0,0 +1 @@
+---
```
kpouget, line 1: I guess you can delete this file

roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_clusterpolicy.yml
```diff
@@ -42,9 +42,20 @@
   retries: 20
   delay: 15
 
+- name: Optionally call the patch_clusterpolicy_role hook to modify the clusterpolicy
+  include_role:
+    name: "{{ patch_clusterpolicy_role }}"
+  when: patch_clusterpolicy_role | default('', true) | trim != ''
```
kpouget, line 48: the issue with this `include_role` design is that it makes it hard to pass parameters to the `patch` role,
eg:
```
toolbox/gpu-operator/set_repo-config.sh REPO_FILE DEST_DIR
toolbox/gpu-operator/set_repo-config.sh --rhel-beta
```
it would be hard to replicate this example without creating a new script or adding many arguments to the original one ...

I'm not sure what's the best way to organize that ....

kpouget: for this task, I think what would be the best from my POV is something like this:
```
default:
  patch_clusterpolicy_role: echo nothing to do for patching

shell: "{{ patch_clusterpolicy_role }} <  "{{ artifact_extra_logs_dir }}"/gpu_operator_clusterpolicy.json"
env: # (not ansible) 
   SRC_CLUSTER_POLICY: "{{ artifact_extra_logs_dir }}"/gpu_operator_clusterpolicy.json"
   DST_CLUSTER_POLICY: "{{ artifact_extra_logs_dir }}"/gpu_operator_clusterpolicy_patched.json"
```
this way, the patching is fully independent from the deployment, and it can receive custom parameters from the invocation command. Mind that we'll have to `unset` some ansible logging variables before calling `source _common.sh` to avoid messing up with the logging directories.

roles/build-operand/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+       | sed 's|{{ '{{' }} build_operand_extra_docker_strategy {{ '}}' }}|{{ build_operand_extra_docker_strategy | to_yaml | trim }}|' \
+       | oc apply -f-
+    args:
+      warn: false # don't warn about using sed here
+
+  - name: Wait for the image to be built
+    shell: oc logs -f "buildconfig/{{ build_operand_imagestream_name }}" -n "{{ build_operand_namespace }}" > /dev/null
+
+- name: Ensure that the image exists
+  command: oc get imagestreamtag -n "{{ build_operand_namespace }}" "{{ build_operand_imagestream_name }}:{{ build_operand_tag }}" -oname --ignore-not-found
```
kpouget, line 52: this `--ignore-not-found` prevents the task from failing

kpouget:
outdated, closing

/close

PR #182, opened 2021-06-02
Support entitlement on no-workers clusters

omertuc:
On no-worker clusters, we want to target the masters with our machine
configs rather than the workers, because there are no workers.

We check that by looking at the Scheduler CR `.spec.mastersSchedulable`
stanza

### DIFF
--- a/roles/entitlement_deploy/files/apply_template.py
+++ b/roles/entitlement_deploy/files/apply_template.py
@@ -1,19 +1,33 @@
-#! /usr/bin/python3
+#!/usr/bin/env python3
 
-import fileinput
 import base64
 import sys
 
-template = sys.argv[1]
-keyname = sys.argv[2]
-filename = sys.argv[3]
-print(f"Replacing '{keyname}' with the content of '{filename}'", file=sys.stderr)
 
-with open(filename, "rb") as f:
-    file_b64 = base64.b64encode(f.read()).decode()
+def main():
+    keyname = sys.argv[1]
+    value = sys.argv[2]
 
-found = False
-for line in fileinput.input(files=[template]):
-    if keyname in line: found = True
-    print(line.replace(keyname, file_b64), end="")
-exit(0 if found else 1)
+    if value.startswith("@"):
+        filename = value[1:]  # Skip @
+        print(
+            f"Replacing '{keyname}' with the base64 of the content of '{filename}'",
+            file=sys.stderr,
+        )
+
+        with open(filename, "rb") as f:
+            value = base64.b64encode(f.read()).decode()
+    else:
+        print(f"Replacing '{keyname}' with '{value}'", file=sys.stderr)
+
+    found = False
+    for line in sys.stdin:
+        if keyname in line:
+            found = True
+        print(line.replace(keyname, value), end="")
+
+    exit(0 if found else 1)
+
+
+if __name__ == "__main__":
+    main()
--- a/roles/entitlement_deploy/tasks/main.yml
+++ b/roles/entitlement_deploy/tasks/main.yml
@@ -14,18 +14,34 @@
 - name: Compute the md5sum of the entitlement files (debug)
   command: md5sum '{{ entitlement_rhsm }}' '{{ entitlement_pem }}'
 
+- name: Get cluster control plane topology
+  command: oc get scheduler -ojsonpath="{.items[0].spec.mastersSchedulable}\n"
+  register: masters_are_also_workers
+
+- name: Set master as machine config target role in single node clusters
+  when: masters_are_also_workers.stdout == "true"
+  set_fact:
+      machine_config_role: master
+
+- name: Set worker as machine config target role in non-single-node clusters
+  when: not masters_are_also_workers.stdout == "false"
+  set_fact:
+      machine_config_role: worker
+
 - name: "Deploy RHSM from file '{{ entitlement_rhsm }}'"
   shell:
     set -o pipefail;
-    python3 "{{ entitlement_py_apply }}"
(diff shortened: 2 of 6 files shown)


roles/entitlement_deploy/files/apply_template.py
```diff
@@ -1,19 +1,33 @@
-#! /usr/bin/python3
+#!/usr/bin/env python3
 
-import fileinput
```
omertuc, line 3: Changed it so it no longer reads from a file, just stdin - so we can stack several apply_template.py calls in pipes

roles/entitlement_deploy/files/apply_template.py
```diff
@@ -1,19 +1,33 @@
...
+def main():
+    keyname = sys.argv[1]
+    value = sys.argv[2]
 
-found = False
-for line in fileinput.input(files=[template]):
-    if keyname in line: found = True
-    print(line.replace(keyname, file_b64), end="")
-exit(0 if found else 1)
+    if value.startswith("@"):
```
omertuc, line 11: Added support for either reading the contents from a file, or simply using the argument passed to it. Because I just want to change a VAR name to "worker" or "master"

roles/entitlement_deploy/tasks/main.yml
```diff
@@ -36,7 +52,8 @@
   - name: "Deploy the repo CA from file '{{ entitlement_mc_rhsm_ca }}'"
     shell:
       set -o pipefail;
-      python3 "{{ entitlement_py_apply }}"
-              "{{ entitlement_mc_rhsm_ca }}" BASE64_ENCODED_RHSM_CA_FILE "{{ entitlement_repo_ca }}"
+      cat "{{ entitlement_mc_rhsm_ca }}"
```
omertuc, line 55: I used `cat` instead of `<` just because it allowed me to align the `python3` calls nicely

roles/entitlement_inspect/tasks/main.yml
```diff
@@ -20,6 +20,10 @@
   command: oc describe MachineConfigPool/worker
   failed_when: false
 
+- name: Get the description of the master MachineConfigPool
+  command: oc describe MachineConfigPool/master
+  failed_when: false
```
kpouget, line 25: can you store the description in `{{ artifact_extra_logs_dir }}mcp_master.descr` (and same for `worker` above)?
that makes the logs easier to read

roles/entitlement_deploy/tasks/main.yml
kpouget: would it be complicated to use Ansible template instead of the Python script?

just wondering what would be simpler/easier to read

roles/entitlement_deploy/files/apply_template.py
kpouget: could you add a line of comment explaining what the script does / how the parameters are used?

I understood after reading the task calling the script, the update makes sense

I'm wondering if we should use Ansible `template` task instead, or not

roles/entitlement_deploy/tasks/main.yml
```diff
@@ -14,18 +14,34 @@
...
+
+- name: Set master as machine config target role in single node clusters
+  when: topology.stdout == "SingleReplica"
+  set_fact:
+      machine_config_role: master
+
+- name: Set worker as machine config target role in non-single-node clusters
+  when: not topology.stdout == "SingleReplica"
+  set_fact:
+      machine_config_role: worker
```
kpouget, line 29: @omertuc could you add a way bypass this?
```
if empty(force_machine_config_role):
  try with the topology
else:
  machine_config_role = force_machine_config_role
```
this `force_machine_config_role` can be set in the toolbox, either now or later when we switch to Python+simpler arg parsing

kpouget: hey @omertuc, it would be nice if you could update this PR with at least this comment addressed,
the rest is nothing important, but this comment would allow entitlement deployment on all-in-one clusters

omertuc: Let me know if the recent push works for you

roles/entitlement_inspect/tasks/main.yml
kpouget: could you write this into `        > {{ artifact_extra_logs_dir }}/master_MachineConfigPool.desc` (don't forget to also change the task from `command` to `shell`)

kpouget:
thanks for the updated @omertuc, I just pointed out a minor fix (write output into a dedicated file instead of log file),
can you then trigger a GPU Operator test before we can merge?

omertuc:
>would it be complicated to use Ansible template instead of the Python script?
just wondering what would be simpler/easier to read

Yeah a template would've definitely made this easier. But since it works, I prefer to not touch it for now.

omertuc:
@kpouget any clue?

```
msg: The task includes an option with an undefined variable. The error was: 'machine_config_role' is undefined
```

https://github.com/openshift-psap/ci-artifacts/blob/5ae18d86a0765cb7dead103e388652f64f87045b/roles/entitlement_deploy/tasks/main.yml#L21-L45

roles/entitlement_deploy/tasks/main.yml
omertuc: answered here https://github.com/openshift-psap/ci-artifacts/pull/182#issuecomment-884958996

kpouget:
Thanks @omertuc for the updates :)

/lgtm
/hold

omertuc:
Why the /hold?

kpouget:
just to let you merge it when you want!

PR #184, opened 2021-06-02
trim extra spaces and empty white lines

ArangoGutierrez:
just a minor cleaning :) 

```bash
find . -path ./.git -prune -false -o -type f | xargs sed s'!\s*$!!' -i
```


kpouget:
thanks Eduardo,
2 minor comments:
- `*.patch` shouldn't be modified, I guess you can add it to the `find` command you're running (just add `! -iname '*.patch'`)
- could you add it as a GitHub Actions test? something like clone/find&fix/fail if git is dirty
does this command handles the EOF/EOL as well?

ArangoGutierrez:
> thanks Eduardo,
> 2 minor comments:
> 
>     * `*.patch` shouldn't be modified, I guess you can add it to the `find` command you're running (just add `! -iname '*.patch'`)
> 
>     * could you add it as a GitHub Actions test? something like clone/find&fix/fail if git is dirty
>       does this command handles the EOF/EOL as well?

the space being removed from the patch file won't affect it's functionality, I don't get why you want to preserve that trailing white space? 

Sure I can do a github action, but maybe not in this PR. I might do that later on, or we could hold this PR

PR #186, opened 2021-06-07
Fix storage of docs/img/ci-dashboard.png

kpouget:
This is necessary so that the image gets always copied to the
generated directory, even if not 'directly' referenced. We use an
absolute path instead, so that the image can be shown from multiple
URLs:
- https://github.com/openshift-psap/ci-artifacts/blob/master/README.rst
- https://github.com/openshift-psap/ci-artifacts/
- https://openshift-psap.github.io/ci-artifacts/intro.html


kpouget:
@jmencak can you have a look and lgtm?

it should get the image correctly displayed in the readme and in the doc

... but only merging and generating will confirm that it's correctly automated!

jmencak:
let'confirm this works then :)
/lgtm

kpouget:
thanks :)
/approve

kpouget:
yes, worked as expected :)

Issue #163, opened 2021-05-19
set_scale.sh: cannot specify the source machineset


kpouget:
Currently, `./toolbox/cluster/set_scale.sh` cannot be customized to decide which `MachineSet` will be used to derive the new MachineSet:

```
- name: Get the names of an existing worker machinesets (of any instance type)
  command:
    oc get machinesets -n openshift-machine-api -o
    jsonpath='{range .items[?(@.spec.template.metadata.labels.machine\.openshift\.io/cluster-api-machine-role=="worker")]}{.metadata.name}{"\n"}{end}'
  register: oc_get_machinesets
  failed_when: not oc_get_machinesets.stdout
```

it would be nice to have the ability to easily override `oc_get_machinesets.stdout` to specify which machineset to use as a base.

Quick and dirty example of what I did to work around that:
```
- name: Get the names of an existing worker machinesets (of any instance type)
  command:
    echo kpouget-20210519-kf6rn-worker-eu-central-1b
  register: oc_get_machinesets
  failed_when: not oc_get_machinesets.stdout
```

The reason for that is that the instance-type I want isn't available in `eu-central-1a`, only in `1b`

kpouget:
\assign @pamann

kpouget:
addressed with
https://github.com/openshift-psap/ci-artifacts/pull/234
https://github.com/openshift-psap/ci-artifacts/pull/243

/close

PR #189, opened 2021-06-08
WIP: specify machineset in set_scale.sh

pamann:
Wasn't sure the best way to do argument validation with multiple optional arguments, so only the required args are validated.


jmencak:
I'd personally vote to move all the options in front of the real arguments as it has been a convention for years.  Nothing against this PR, the issue (options after arguments) was introduced prior to it. Options before arguments would simplify the parsing and you could then use something like this to process them.

```
while true
do
    case "${1:-}" in
        -f|--force)
            force=Y
            ;;
        -a|--another_option)
            another_option=Y
            ;;
        -*) echo "$0: invalid option '${1:-}'" >&2
            usage 1
            ;;
        *)  break 
            ;;
    esac
    shift
done

my_arg1=${1}
my_arg2=${2} # or just use shift here
```
@kpouget WDYT?

kpouget:
@jmencak yes, makes sense
we really need to define a clean&common way to pass and parse arguments :#

@pamann, I think there is something off with the current PR:
```
 USAGE: ${0} <instance_type> <scale> [machineset] [--force]
```
reading the code, I think `machineset` is a number, like `0` for the 1st machineset, `1` for the second, etc, is that correct?

if yes, that's not the easiest way to go for the user, I would prefer:
```
BASE_MACHINESET=kpouget-20210519-kf6rn-worker-eu-central-1b
./toolbox/cluster/set_scale.sh g4dn.xlarge 1 $BASE_MACHINESET
# or maybe
./toolbox/cluster/set_scale.sh g4dn.xlarge 1 --base-machineset $BASE_MACHINESET
```
where `$BASE_MACHINESET` is the name of a machineset:
```
$ oc get machineset/$BASE_MACHINESET -n openshift-machine-api
...
$ oc get machinesets -n openshift-machine-api -oname 
...
```

you can also add an ansible test task that fails if the machineset doesn't exist
```
name: Ensure that the base machineset exists
command: oc get machinesets/{{ base_machineset }} -oname
```

Give it a try this afternoon, and if not clear we'll see tomorrow morning (your time) how to get that working :)

omertuc:
> we really need to define a clean&common way to pass and parse arguments :#

I have a branch in progress I'm working on from time to time to transform the toolbox to Fire...

kpouget:
> > we really need to define a clean&common way to pass and parse arguments :#
> 
> I have a branch in progress I'm working on from time to time to transform the toolbox to Fire... The commands themselves are trivial, _common.sh is turning out to be a bit of a pain.

nice, feel free to share even a WIP

kpouget:
@omertuc can you have a look and provide some feedback / guidance to @pamann?

roles/cluster_ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -6,9 +6,20 @@
   register: oc_get_machinesets
   failed_when: not oc_get_machinesets.stdout
 
+- name: Set base_machineset
+  command:
+    echo {{ oc_get_machinesets.stdout[0] }}
+  register: base_machineset
+  when: not base_machineset
```
omertuc, line 13: Please declare `base_machineset` in the [`defaults/main/config.yml` file](https://github.com/openshift-psap/ci-artifacts/blob/8b6bb0136d21d333367b12c33d05375503ef60e8/roles/cluster_set_scale/defaults/main/config.yml#L1)

toolbox/cluster/set_scale.sh
```diff
@@ -36,18 +35,30 @@ fi
...
+        BASE_MACHINESET=${1}
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e base_machineset=${BASE_MACHINESET}"
+        shift
+        if [ $# -ne 0 && "${1}" == "--force" ]; then
+            ANSIBLE_OPTS="${ANSIBLE_OPTS} -e force_scale=true"
+        fi
+    elif ["${1}" == "--force" ]; then
+        echo "Setting cluster ${INSTANCE_TYPE} machinesets to have scale '${SCALE}'"
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e force_scale=true"
+    fi
```
kpouget, line 53: here you can simplify a bit by removing the duplicated `if` block:
```
# pseudo bash
if $1 == --base-machineset; then
       shift
       BASE_MACHINESET=${1}
       ANSIBLE_OPTS=...
       shift
fi

if ${1:-} == --force; then
    echo ... 
    ....
fi
```

not the bash notation `${1:-}` which gives the value of `$1` *if it is set, otherwise the empty default value. 
General form is `${var:-default_value}` which `$var` or `default_value`.
Very useful when `set -o nounset` (`set -u`) is defined!

toolbox/cluster/set_scale.sh
```diff
@@ -36,18 +35,30 @@ fi
...
+    if [ "${1}" == "--base-machineset" ]; then
+        shift
+        BASE_MACHINESET=${1}
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e base_machineset=${BASE_MACHINESET}"
+        shift
+        if [ $# -ne 0 && "${1}" == "--force" ]; then
+            ANSIBLE_OPTS="${ANSIBLE_OPTS} -e force_scale=true"
+        fi
+    elif ["${1}" == "--force" ]; then
+        echo "Setting cluster ${INSTANCE_TYPE} machinesets to have scale '${SCALE}'"
```
kpouget, line 51: this comment can be removed, it's already printed near the end ot the script

toolbox/cluster/set_scale.sh
```diff
@@ -36,18 +35,30 @@ fi
...
+shift 
+SCALE=${1}
 shift
+if [ $# -ne 0 ]; then
+    if [ "${1}" == "--base-machineset" ]; then
+        shift
+        BASE_MACHINESET=${1}
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e base_machineset=${BASE_MACHINESET}"
+        shift
+        if [ $# -ne 0 && "${1}" == "--force" ]; then
```
kpouget, line 47: this `and` isn't valid, see there for examples https://stackoverflow.com/questions/13408493/an-and-operator-for-an-if-statement-in-bash using `[`

or replace `[` by `[[` (and closing)

toolbox/cluster/set_scale.sh
```diff
@@ -36,18 +35,30 @@ fi
...
-ANSIBLE_OPTS="${ANSIBLE_OPTS} -e machineset_instance_type=${INSTANCE_TYPE}"
-
+shift 
+SCALE=${1}
 shift
+if [ $# -ne 0 ]; then
+    if [ "${1}" == "--base-machineset" ]; then
+        shift
+        BASE_MACHINESET=${1}
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e base_machineset=${BASE_MACHINESET}"
```
kpouget, line 45: you can add an `echo` here, `Setting base machineset to '${BASE_MACHINESET}'`

toolbox/cluster/set_scale.sh
```diff
@@ -36,18 +35,30 @@ fi
...
 INSTANCE_TYPE=${1}
-ANSIBLE_OPTS="${ANSIBLE_OPTS} -e machineset_instance_type=${INSTANCE_TYPE}"
-
+shift 
+SCALE=${1}
 shift
+if [ $# -ne 0 ]; then
+    if [ "${1}" == "--base-machineset" ]; then
+        shift
+        BASE_MACHINESET=${1}
```
kpouget, line 44: this fails if the base-machineset value isn't set:
```
./toolbox/cluster/set_scale.sh: line 44: 1: unbound variable
```
can you add a `-z ${1:-}`check + message + exit if missing?
these manual and adhoc checks will go away once we move to Python, but meanwhile .... !

kpouget (changes requested):
Hello Pam, I suggested a few things to improve,
and as we discussed, we can add a check in the ansible playbook to validate that the `machineset` exists, and print a message otherwise,
Currently, I got:
```---
roles/cluster_ensure_machineset/tasks/derive_machineset.yml:20
TASK: cluster_ensure_machineset : Store the MachineSet json definition in a file
----- FAILED ----
msg: non-zero return code

<command> oc get 'machineset/kpouget-20210609-tfxv9-worker-eu-central-toto' -n openshift-machine-api -o json > /tmp/ci-artifacts_20210614/023__cluster__set_scale/source_machineset.json

<stderr> Error from server (NotFound): machinesets.machine.openshift.io "kpouget-20210609-tfxv9-worker-eu-central-toto" not found
----- FAILED ----
```
which is not so bad, but a check&fail would make it clear that the fault is on the user provided parameter, and not an internal problem

toolbox/cluster/set_scale.sh
kpouget: this `echo` is still duplicated (see line 59)

kpouget: (actually it seems it was already duplicated in the original version)

roles/cluster_ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -6,9 +6,22 @@
...
+- name: Check if base_machineset specified
+  set_fact:
+    base_machineset={{ base_machineset | default(oc_get_machinesets.stdout_lines[0]) }}
+
+- name: Ensure that the specified base machineset exists
+  fail: msg="Specified base machineset ( {{ base_machineset }} ) is invalid or does not exist"
+  when: base_machineset not in oc_get_machinesets.stdout_lines
+
+- name: Ensure that the base machineset exists
+  command: oc get machinesets/{{ base_machineset }} -oname
```
kpouget, line 18: you need to specify the namespace here `-n openshift-machine-api`, otherwise it will only work when you're in the `openshift-machine-api` namespace

roles/cluster_ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -6,9 +6,22 @@
   register: oc_get_machinesets
```
kpouget, line 6:
```
roles/cluster_ensure_machineset/tasks/main.yml:15
TASK: cluster_ensure_machineset : Create the new machineset
  - include:	derive_machineset.yml
included: /home/kevin/openshift/ci-artifacts/roles/cluster_ensure_machineset/tasks/derive_machineset.yml for localhost
```
could you change the `name` there to: `Derive the new machineset`? the `Create` is a bit confusing here, sounds like a simple/single task

roles/cluster_ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -6,9 +6,22 @@
...
   failed_when: not oc_get_machinesets.stdout
 
+- name: Check if base_machineset specified
```
kpouget, line 9: can you change the title to:
> "Check if base_machineset specified: {{ base_machineset | default('not set') }}"

toolbox/cluster/set_scale.sh
```diff
@@ -36,18 +35,27 @@ fi
...
-    ANSIBLE_OPTS="${ANSIBLE_OPTS} -e force_scale=true"
+if [ $# -ne 0 ]; then
+    if [ ${1:-} == "--base-machineset" ]; then
+        shift
+        BASE_MACHINESET=${1:-}
+        echo "Setting base machineset to '${BASE_MACHINESET}'"
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e base_machineset=${BASE_MACHINESET}"
+        shift
+    fi
+    if ["${1:-}" == "--force" ]; then
```
kpouget, line 49: space missing here: `[ "${1:-}"`

roles/cluster_ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -6,9 +6,22 @@
...
+  set_fact:
+    base_machineset={{ base_machineset | default(oc_get_machinesets.stdout_lines[0]) }}
+
+- name: Ensure that the specified base machineset exists
+  fail: msg="Specified base machineset ( {{ base_machineset }} ) is invalid or does not exist"
```
kpouget, line 14: could you remove the spaces inside the parentheses? (`({{ base_machineset }})`)

roles/cluster_ensure_machineset/tasks/derive_machineset.yml
```diff
@@ -6,9 +6,22 @@
   register: oc_get_machinesets
   failed_when: not oc_get_machinesets.stdout
 
+- name: Check if base_machineset specified
+  set_fact:
+    base_machineset={{ base_machineset | default(oc_get_machinesets.stdout_lines[0]) }}
```
kpouget, line 11: can you add `base_machineset` in `defaults/main/config.yml`:
```
# Optional: name of a machineset to use to derive the new one
# If empty, using the first machineset found in the listing
base_machineset: ""
```

kpouget (commented):
thanks @pamann, it seems to work well with a few details to fix

... I spent quite a while though trying to debug it ... because I was calling the script wrongly:
```
toolbox/cluster/set_scale.sh g4dn.xlarge 1 --base-machineset kpouget-20210625-svffk-worker-eu-central-1c
```
 and of course my region `1c` was never used 🤦🏻

kpouget:
ping @pamann , any update on this one?

kpouget:
@pamann seems to look good :)
only the yaml lint test if failing because of the missing new-line at the end of `roles/cluster_ensure_machineset/tasks/main.yml`
and could you add an entry in the [changelog](https://github.com/openshift-psap/ci-artifacts/blob/master/CHANGELOG.rst) and in the documentation [page](https://github.com/openshift-psap/ci-artifacts/blob/master/docs/toolbox/cluster.rst)?

@omertuc would you be able to give this PR a try (on monday)?
specify eg the machineset zone c, and make sure the new machines are created from this machineset (this is the thing I wanted to solve with [this](https://github.com/openshift-psap/ci-artifacts/issues/163) ticket)

kpouget:
closing this PR as @pamann won't be able to pursue the work

I've rebased the code from this PR on top of `master` in https://github.com/openshift-psap/ci-artifacts/pull/234
I didn't have to change anything for the Ansible part, the only unsolved points where in `set_scale.sh`, and this file disappeared as part of the rebase (move to Python)

PR #190, opened 2021-06-13
WIP CI using mirror

omertuc:


build/root/usr/local/bin/entitle.sh
```diff
@@ -65,8 +66,8 @@ if [ ! -e "$ENTITLEMENT_PEM" ]; then
     exit 1
 fi
 
-echo "INFO: Deploying the entitlement with PEM key from ${ENTITLEMENT_PEM}"
-toolbox/entitlement/deploy.sh --pem "${ENTITLEMENT_PEM}"
+echo "INFO: Deploying the entitlement with PEM key from ${ENTITLED_MIRROR_CLIENT_CREDS}"
+toolbox/entitlement/deploy.sh --yum-client-auth="${ENTITLED_MIRROR_CLIENT_CREDS}"
```
omertuc, line 70: Temporary change, just to make sure it works

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
```diff
@@ -46,6 +48,8 @@ collect_must_gather() {
 validate_gpu_operator_deployment() {
     trap collect_must_gather EXIT
 
+    toolbox/gpu-operator/set_repo-config.sh "${SCRIPT_DIR}"/ci_mirror.repo
```
kpouget, line 51: `build/root/usr/local/bin/ci_mirror.repo` isn't the right location for the repolist, maybe `build/resources`?

I would like to change (later, maybe when introducing the cli parsing framework) `build` to something more meaningful, maybe `ci_image`? (cc @ArangoGutierrez )

roles/gpu_operator_wait_deployment/tasks/main.yml
```diff
@@ -65,7 +65,7 @@
     register: has_gpu_feature_discovery_labels
     until:
     - has_gpu_feature_discovery_labels.stdout != ""
-    retries: 10
+    retries: 30
```
kpouget, line 68: not directly related, did you have issue with this test?

toolbox/entitlement/deploy.sh
```diff
@@ -1,41 +1,66 @@
...
+        ENTITLEMENT_PEM="${value}"
+  elif [[ "${flag}" == "--yum-client-auth" ]]; then
+        YUM_CLIENT_AUTH_FILE="${value}"
+  fi
+
+  shift
+done
+
+if [[ "$ENTITLEMENT_PEM" ]] && [[ "$YUM_CLIENT_AUTH_FILE" ]]; then
+  echo "FATAL: Having both entitlement and yum client auth certificates is currently unsupported"
```
kpouget, line 41: that's not what the `usage` text says,
I guess it makes sense to keep them separated, as you don't normally want to deploy an entitlement *and* a yum auth key

toolbox/entitlement/deploy.sh
```diff
@@ -1,41 +1,66 @@
...
 
-if [[ "$1" == "--pem" ]]; then
-    ANSIBLE_OPTS="${ANSIBLE_OPTS} -e entitlement_pem=$(realpath $2)"
-    echo "Using '$2' as PEM key"
-else
-    echo "ERROR: please pass a valid flag."
-    usage
-    exit 1
+if [[ -z "$ENTITLEMENT_PEM" ]] && [[ -z "$YUM_CLIENT_AUTH_FILE" ]]; then
+  echo "FATAL: Please provide at-least one flag"
```
kpouget, line 47: I don't think the `-` belongs there (in `at-least`)

toolbox/entitlement/deploy.sh
```diff
@@ -1,41 +1,66 @@
...
-    ANSIBLE_OPTS="${ANSIBLE_OPTS} -e entitlement_repo_ca=$(realpath $4)"
-    echo "Using '$4' as repo CA"
+if [[ "$YUM_CLIENT_AUTH_FILE" ]]; then
+  ANSIBLE_OPTS="${ANSIBLE_OPTS} -e entitlement_repo_yum_client_auth_pem_file=${YUM_CLIENT_AUTH_FILE}"
+  echo "Using '$YUM_CLIENT_AUTH_FILE' as yum client auth credentials"
 fi
 
+
+CURR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+source ${CURR_DIR}/../_common.sh
```
kpouget, line 64: can you change it to `THIS_DIR`, for consistency. I wonder why this one used a different name .... !

kpouget (changes requested):
thanks @omertuc, I added some comments,
and we'll have to decide how we want to test it!

build/root/usr/local/bin/entitle.sh
```diff
@@ -65,10 +66,10 @@ if [ ! -e "$ENTITLEMENT_PEM" ]; then
...
 
-if ! toolbox/entitlement/wait.sh; then
-    echo "FATAL: Failed to properly entitle the cluster, cannot continue."
-    exit 1
-fi
```
kpouget, line 74: can you get this test to work? it may not be straightforward, but it would allow catching mirror errors earlier than in the driver container not deploying ...
or add a GPU Operator test to detect when the driver fails to build

build/root/usr/local/bin/unentitle.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash -e
+
+set -o pipefail
+set -o errexit
+set -o nounset
+set -x
+
+toolbox/entitlement/undeploy.sh
```
kpouget, line 8: why not calling ` toolbox/entitlement/undeploy.sh ` directly?

and I think you'll need to wait for the cluster to be unentitled, one way or another
one way could be to scale down to `0` the # of GPU nodes, unentitle, then scale up to `1`,
this way you know that the GPU node will come up without the entitlement files


another way is to pin the testing pod to nodes with `nvidia.com/gpu.present=true`, and wait until the entitlement test fails

kpouget: (this file is never called BTW)

roles/gpu_operator_wait_deployment/tasks/main.yml
omertuc: Yes. This failed for me. It looked like the driver build took a bit longer due to the mirror, not sure.

build/root/usr/local/bin/entitle.sh
omertuc: Good point

build/root/usr/local/bin/unentitle.sh
omertuc: > why not calling toolbox/entitlement/undeploy.sh directly?

Wasn't sure where the toolbox is mapped exactly relative to the CI working directory

>one way could be to scale down to 0 the # of GPU nodes, unentitle, then scale up to 1,
this way you know that the GPU node will come up without the entitlement files
Yeah I guess that if I do that I don't have to use undeploy at all

kpouget: > Wasn't sure where the toolbox is mapped exactly relative to the CI working directory

yes, there are things unclear in this regard, but the implicit assumption is that `run` and `ci-entrypoint_*` all run with `CWD=$HOME`, never change directory, and `toolbox/...` is available from there

omertuc: > and I think you'll need to wait for the cluster to be unentitled, one way or another
Maybe `toolbox/entitlement/undeploy.sh` playbook should do this waiting. What do you think?

kpouget: the thing is that it's hard to know how long to wait, eg,
a) do you wait for all the nodes to be rebooted? 
b) do you wait for one random node to be rebooted (that what we do for `toolbox/entitlement/wait.sh` as my goal is to validate that the entitlement is working, not to wait for all the nodes to be entitled)

maybe  `toolbox/entitlement/wait.sh` could be easily updated to wait for a pod failing the entitlement test (case `b)` above,
but that's not a full solution
or maybe it can also receive a node name, to specify which node we want to test
```
toolbox/entitlement/wait.sh --not-entitled --node=my_node
```
(mind that node "name != node hostname" for the node selector)

omertuc: You can wait for machineconfigpool status:

After removing machineconfig:
image

After the removal is finally complete:
image

kpouget: @omertuc yes indeed that can be an easy way to wait for a full removal of the the MachineConfigs,
then it can become a flag of `undeploy.sh`:
```
toolbox/entitlement/undeploy.sh --wait
```

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
```diff
@@ -4,11 +4,16 @@ set -o pipefail
...
+
 prepare_cluster_for_gpu_operator() {
     trap collect_must_gather ERR
 
     toolbox/cluster/capture_environment.sh
-    entitle.sh
+
+    if [[ ${USE_ENTITLED_MIRROR} != "1" ]]; then
+        entitle.sh --mirror
+    fi
```
kpouget, line 16: not sure to understand this block, `if not use-the-entitled-mirror; then entitle.sh --mirror` ?

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
```diff
@@ -104,7 +113,14 @@ if [ -z "${1:-}" ]; then
...
 
 action="$1"
-shift
+shift || true
+
+if [[ "${1:-}" == "--mirror" ]]; then
+    USE_ENTITLED_MIRROR=1
+else
+    USE_ENTITLED_MIRROR=0
+fi
```
kpouget, line 122: what about a one-liner?
```
USE_ENTITLED_MIRROR=$([[ "${1:-}" == "--mirror" ]] && echo 1 || echo 0)
#or
[[ ! "${1:-}" = "--mirror" ]]; USE_ENTITLED_MIRROR=$? 
```

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
omertuc: Just old unpushed code, was focusing on pushing to `release-next`.... updated now

CHANGELOG.rst
```diff
@@ -22,6 +22,14 @@ Retro-compatibility breaks
 
 - Add nfd test_master_branch protocol `#179 https://github.com/openshift-psap/ci-artifacts/pull/179>`_
 
+- Changes to ``toolbox/entitlement/deploy.sh`` to allow deploying yum client auth credentials machine configs instead of entitlement machineconfigs
+
+  - ``toolbox/entitlement/deploy.sh --pem /path/to/key.pem`` will now require an equals sign, e.g. ``toolbox/entitlement/deploy.sh --pem=/path/to/key.pem``
+  - Added new flag ``toolbox/entitlement/deploy.sh --yum-client-auth=/path/to/credentials.pem`` to allow deploying PEM credentials as machineconfigs
+
+...
+>>>>>>> 7590d9d... CI using mirror
```
kpouget, line 31: rebase garbage ;-)

build/root/usr/local/bin/ci_mirror.repo
```diff
@@ -0,0 +1,51 @@
+[rhocp-4.5-for-rhel-8-x86_64-rpms]
```
kpouget, line 1: can you move this file out of `build/root/usr/local/bin`? maybe `build/root/etc` to start with (in the short term I want to reorganize `build/` into more "logical names"that `usr/...` )

kpouget: it could be that it becomes a new `subproject`, WDYT? (cc @ArangoGutierrez )

build/root/usr/local/bin/entitle.sh
```diff
@@ -29,12 +29,19 @@ if toolbox/entitlement/test_cluster.sh --no-inspect; then
     exit 0
 fi
 
+if [[ "${1:-}" == "--mirror" ]]; then
+    USE_ENTITLED_MIRROR=1
+else
+    USE_ENTITLED_MIRROR=0
+fi
```
kpouget, line 36: one liner?

roles/entitlement_deploy/files/mc_yum_client_auth.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: worker
+  name: 50-yum-client-auth
+spec:
+  config:
+    ignition:
+      version: 2.2.0
```
kpouget, line 10: I think I used `3.0.0` for the other files

toolbox/entitlement/undeploy.sh
```diff
@@ -5,3 +5,5 @@ CURR_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 source ${CURR_DIR}/../_common.sh
 
 exec ansible-playbook ${ANSIBLE_OPTS} playbooks/entitlement_undeploy.yml
+
+sleep 600
```
kpouget, line 9: this command is never executed, because of the `exec` above

kpouget (commented):
I added a few comments, let me know when it's ready for testing

PR #195, opened 2021-06-16
nfd_operator_deploy_from_operatorhub: fix broken path to files

kpouget:

PR #196, opened 2021-06-16
Introduce a Github Action for checking ansible variable consistency

kpouget:
This PR creates a new `pull_request` GitHub Action to validate:
1. that all the variables defining a filepath (`roles/`) do point to an existing file
2. that all the variables defined are actually used in their role (with an exception for symlinks, shared between `undeploy` -> `deploy`).

This should prevent the problem solved by https://github.com/openshift-psap/ci-artifacts/pull/195 to happen again.

The PR also fixed problems found by the checks (only unused variables).

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,6 +23,7 @@ Red Hat PSAP CI-Artifacts toolbox
    toolbox/nto
    toolbox/sro
    toolbox/local-ci
+   toolbox/repo
 
 
 Documentation generated on |today| from |release|.
--- a/toolbox/repo/validate_role_files.py
+++ b/toolbox/repo/validate_role_files.py
@@ -1,0 +1,80 @@
+#! /usr/bin/env python
+
+# This script ensures that all the Ansible variables defining a
+# filepath (`roles/`) do point to an existing file.
+
+import glob, os, sys
+import yaml
+import pathlib
+
+THIS_DIR = pathlib.Path(__file__).absolute().parent
+TOP_DIR = THIS_DIR.parent.parent
+# go to top directory
+os.chdir(TOP_DIR)
+
+ROLES_VARS_GLOB = "roles/*/vars/*/*"
+FILE_PREFIX = "roles/"
+
+def validate_role_vars_files(filename, yaml_doc):
+    errors = 0
+
+    for key, value in yaml_doc.items():
+        try:
+
+            if not value.startswith(FILE_PREFIX):
+                print(f"{key}: {value} --> not starting with '{FILE_PREFIX}', ignoring.")
+                continue
+        except AttributeError: # value.startswith
+            print(f"{key}: --> no a string, ignoring.")
+            continue
+
+        if not pathlib.Path(value).exists():
+            errors += 1
+
+            print(f"ERROR: {key}: {value} --> not found")
+            continue
+
+        # file found, nothing to do
+
+    return errors
+
+def traverse_role_vars():
+    errors = 0
+    for filename in TOP_DIR.glob(ROLES_VARS_GLOB):
+        print()
+        print("###", filename.relative_to(TOP_DIR))
+        with open(filename) as f:
+            try:
+                yaml_doc = yaml.safe_load(f)
+            except yaml.YAMLError as e:
+                print(f"--> invalid YAML file ({e})")
+                continue
+
+        if yaml_doc is None:
+            print(f"--> empty file")
+            continue
+
+        file_errors = validate_role_vars_files(filename, yaml_doc)
+        errors += file_errors
+
+        print(f"--> found {file_errors} error{'s' if file_errors > 1 else ''} in {filename.relative_to(TOP_DIR)}")
+
+    return errors
+
+def main():
+    if "-h" in sys.argv or "--help" in sys.argv:
+        print("""\
+This script ensures that all the Ansible variables defining a
+filepath (`roles/`) do point to an existing file.""")
+        return 0
+
+    print(f"INFO: Searching for missing files in '{ROLES_VARS_GLOB}'")
+    errors = traverse_role_vars()
+
+    print()
+    print(f"{'ERROR' if errors else 'INFO'}: found {errors} missing file{'s' if errors > 1 else ''}")
+
+    return 1 if errors else 0
+
+if __name__ == "__main__":
+    exit(main())
--- a/toolbox/repo/validate_role_vars_used.py
+++ b/toolbox/repo/validate_role_vars_used.py
@@ -1,0 +1,98 @@
+#! /usr/bin/env python
+
+# This script ensures that all the Ansible variables defined are
+# actually used in their role (with an exception for symlinks)
+
+import glob, os, sys
+import yaml
+import subprocess
+import pathlib
+import itertools
+import shlex
+
+THIS_DIR = pathlib.Path(__file__).absolute().parent
+TOP_DIR = THIS_DIR.parent.parent
+# go to top directory
+os.chdir(TOP_DIR)
+
+ROLES_VARS_GLOB = "roles/*/vars/*/*"
+ROLES_DEFAULTS_GLOB = "roles/*/defaults/*/*"
+
+def validate_role_vars_used(filename, yaml_doc):
+    errors = 0
+
+    # filename.parts[len(TOP_DIR.parts)] is 'roles'
+    role_name = filename.parts[len(TOP_DIR.parts)+1]
+    for key in yaml_doc:
+        grep_command = ["grep", key,
+                        str(pathlib.Path("roles") / role_name),
+                        "--recursive",
+                        ]
+        proc = subprocess.run(grep_command, capture_output=True)
+
+        if proc.returncode != 0:
+            print("DEBUG:", shlex.join(grep_command))
+            # grep should always find 'key' inside 'filename',
+            # I couldn't find how to exclude one specific file with
+            # grep options....
+            print(proc.stderr)
+            print("ERROR: grep shouldn't have failed ...")
+            proc.check_returncode() # raises a subprocess.CalledProcessError
+
+        count = len(proc.stdout.decode('utf-8').splitlines())
+        count -= 1 # exclude 'key' found inside 'filename'
+        if count == 0:
+            print(f"ERROR: '{key}' not used in role '{role_name}'")
+            errors += 1
+            continue
+
+        # key is used, nothing to do
+
+    return errors
+
+def traverse_role_vars_defaults():
+    errors = 0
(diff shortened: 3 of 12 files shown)


toolbox/repo/validate_role_files.py
```diff
@@ -0,0 +1,68 @@
...
+            print(f"ERROR: {key}: {value} --> not found")
+            continue
+
+        # file found, nothing to do
+
+    return errors
+
+def traverse_role_vars():
+    errors = 0
+    for filename in glob.glob(ROLES_VARS_GLOB):
```
omertuc, line 40: nit: If you're already importing pathlib, you can use its built-in globing functionality like ```python
for path in pathlib.Path(FILE_PREFIX).glob("*/vars/*/*"):
```

File: toolbox/repo/validate_role_files.py [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+        file_errors = validate_role_vars_files(filename, yaml_doc)
+        errors += file_errors
+
+        print(f"--> found {file_errors} error{'s' if file_errors > 1 else ''} in {filename}")
+
+    return errors
+
+if __name__ == "__main__":
+    print(f"INFO: Searching for missing files in '{ROLES_VARS_GLOB}'")
+    errors = traverse_role_vars()
```
omertuc, line 63: nit: It's best to avoid writing code directly under the `if __name...` conditional because variables defined here are in global scope, because they're not under any function.

Usually it's better to `def main():` and then put all this code inside it, then call `main()` from `if __name__ ==...`

toolbox/repo/validate_role_files.py
```diff
@@ -0,0 +1,68 @@
+#! /usr/bin/env python
+
+import glob, os
+import yaml
+import pathlib
+
+THIS_DIR = pathlib.Path(__file__).absolute().parent
```
omertuc, line 7: nit: Maybe call this `SCRIPT_DIR` because `THIS_DIR` might potentially be confused with "Current directory"

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+
+import glob, os
+import yaml
+import subprocess
+import pathlib
+
+THIS_DIR = pathlib.Path(__file__).absolute().parent
+
+# go to top directory
+os.chdir(os.path.join(THIS_DIR, "..", ".."))
```
omertuc, line 11: Inconsistent with the other script (other script was nicer IMO)

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
+#! /usr/bin/env python
+
```
omertuc, line 2: Right now the only documentation for what this script does is in the changelog. I think it would be helpful to add a comment to the top of the file that explains in short what it does.

Same goes for the other script

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+# go to top directory
+os.chdir(os.path.join(THIS_DIR, "..", ".."))
+
+ROLES_VARS_GLOB = "roles/*/vars/*/*"
+ROLES_DEFAULTS_GLOB = "roles/*/defaults/*/*"
+
+def validate_role_vars_used(filename, yaml_doc):
+    errors = 0
+
+    role_name = filename.split("/", 2)[1]
```
omertuc, line 19: Does the `, 2` in the split make any difference in this case?

omertuc: Also, you can use `pathlib.Path(filename).parts` to get a list of path components you can index into

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+    for key in yaml_doc.keys():
```
omertuc, line 20: nit: FYI you can omit the `keys()` call, Python dicts can be iterated directly `for key in yaml_doc:` for keys

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+        proc = subprocess.run(["grep", key, f"roles/{role_name}",
```
omertuc, line 21: You can use `pathlib.Path("roles") / role_name` (division operator) to join path components in a more robust/safe way that string formatting

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+                               "--recursive",
+                               ], capture_output=True)
+
+        if proc.returncode != 0:
+            # grep should always find 'key' inside 'filename'
+            # I couldn't find how to exclude one this file with grep options.
```
omertuc, line 27: Looks like this sentence got messed up:
`# I couldn't find how to exclude one this file with grep options.`
What did you mean?

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+            print(proc.stderr)
+            print("ERROR: grep shouldn't have failed ...")
```
omertuc, line 29: Maybe put the subprocess command arguments list in a variable and include it in this print statement, this way people can copy that command and run it manually to try to better understand what went wrong

toolbox/repo/validate_role_vars_used.py
kpouget: I meant: I couldn't find how to exclude one file with grep options. I wanted to do this:
```
KEY=entitlement_rhsm
ROLE=roles/entitlement_deploy
DEF_FILE=roles/entitlement_deploy/vars/main/resources.yml
grep $KEY $ROLE --exclude=DEF_FILE -r
```
but this doesn't work, this line still appears in the output:
```
roles/entitlement_deploy/vars/main/resources.yml:entitlement_rhsm: 'roles/entitlement_deploy/files/rhsm.conf'
```
(just in case you know how to do it with `grep` only)

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+            proc.check_returncode() # raises a subprocess.CalledProcessError
+
+        count = proc.stdout.count(b'\n')
```
omertuc, line 32: nit: 
```python
count = len(proc.stdout.decode('utf-8').splitlines())
```
for better unicode support (obviously not important here, it's a nit)

toolbox/repo/validate_role_vars_used.py
```diff
@@ -0,0 +1,78 @@
...
+        count -= 1 # exclude 'key' found inside 'filename'
```
omertuc, line 33: I guess https://github.com/openshift-psap/ci-artifacts/pull/196/files#r652538672 was talking about excluding it from the get-go

toolbox/repo/validate_role_vars_used.py
omertuc: I see. Looks like grep `--exclude` is for file name patterns, and those patterns can't include directories. Useless.

kpouget: yes exactly, `--exclude=resources.yml` worked as expected, but that may exclude too many files!

docs/index.rst
```diff
@@ -23,6 +23,7 @@ Red Hat PSAP CI-Artifacts toolbox
    toolbox/nto
    toolbox/sro
    toolbox/local-ci
+   toolbox/repo
```
kpouget, line 26: hum, forgot to add the file :#

toolbox/repo/validate_role_files.py
kpouget: I use this name in the rest of the toolbox, so I'll keep the same name here for consistency

toolbox/repo/validate_role_files.py
kpouget: good point for the shadowing, I didn't think about that!

toolbox/repo/validate_role_vars_used.py
kpouget: ack, I'll even add a `-h` and `--help`!

toolbox/repo/validate_role_vars_used.py
kpouget: 👍🏻 I remembered seing a Py file of yours using it after writing the scripts; I didn't know this library and it's quite nice :)

toolbox/repo/validate_role_vars_used.py
kpouget: using `Path` everywhere, it now gives `filename.parts[len(TOP_DIR.parts)+1]` :)

toolbox/repo/validate_role_vars_used.py
kpouget: I like it, it reads weird the first time, but when you notice you're not trying to "divide strings", it makes sense!

toolbox/repo/validate_role_vars_used.py
kpouget: +1 for printing the command, 
but not sure what you mean with
>  Not just here, might be helpful to print it in case of any error in this role

if I always print it, the output will be very verbose

toolbox/repo/validate_role_vars_used.py
kpouget: I hesitated between binary and utf-8, I thought "if the string can't be converted to utf8 it will fail",
but better assume that utf8 always works, and `b'\n'` might be less generic

toolbox/repo/validate_role_files.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+
+            if not value.startswith(FILE_PREFIX):
+                print(f"{key}: {value} --> not starting with '{FILE_PREFIX}', ignoring.")
+                continue
+        except AttributeError: # value.startswith
+            print(f"{key}: --> no a string, ignoring.")
+            continue
+
+        if not pathlib.Path(value).exists:
```
kpouget, line 31: hum @omertuc can you spot the stupid mistake?

toolbox/repo/validate_role_vars_used.py
omertuc: There's another error path - `if count = 0:` - it should be there too

kpouget: if `count == 0` (0 lines of stdout), `grep` return `1`, right?

omertuc: Right

jmencak:
Thank you for the PR, Kevin.  I've tested this and I also believe that most of the reviewers comments were addressed.  This PR does what's in the description and I do not have anything else to add.
/lgtm

kpouget:
thanks for the review Jiri, appreciated :)

I ran a test before merging, just in case :)

PR #199, opened 2021-06-21
GPU Operator: build and deploy commit from bundle

kpouget:
This PR rewrites the role `gpu_operator_deploy_custom_commit`
(that was based on the GPU Operator helm charts) and turn it into
`gpu_operator_bundle_from_commit`, which builds the operator (same as
before) from the given commit, but then prepares a bundle image.

I wanted to store this bundle image inside the cluster, so that this
role can run standalone, but it turned out to be impossible at the
moment, because 1/ `operator-sdk` fails to deploy from an ImageStream,
2/ the image must be available locally for the `operator-sdk` process
to access it, but also cluster-wide, so that a `catalog` can be
created to reference it (so a pod-local registry doesn't work).

Because of these constraints that I couldn't all solve, I decided to
rely a quay.io image registry, and hence the script requires quay.io
push credentials to work.

Note that contrary to the original script, this one only builds a
bundle (with the operator image pointing inside the
cluster). `toolbox/gpu-operator/deploy_from_operatorhub.sh` must be
called afterwards, pointing to the quay.io bundle image.


kpouget:
> /test gpu-operator-e2e
> /test gpu-operator-commit

both passed \o/

@omertuc @dagrayvid ready for review when you'll have time

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
```diff
@@ -58,17 +58,35 @@ test_master_branch() {
...
+
+    echo "Using Git repository ${gpu_operator_git_repo} with ref ${gpu_operator_git_ref}"
 
     prepare_cluster_for_gpu_operator
-    toolbox/gpu-operator/deploy_from_commit.sh "${CI_IMAGE_GPU_COMMIT_CI_REPO}" \
-                                               "${CI_IMAGE_GPU_COMMIT_CI_REF}" \
+
+    if [ -z "${GPU_OPERATOR_QUAY_BUNDLE_PUSH_SECRET:-}" ]; then
+        GPU_OPERATOR_QUAY_BUNDLE_PUSH_SECRET=/var/run/psap-entitlement-secret/openshift-psap-openshift-ci-secret.yml
+    fi
```
dagrayvid, line 76:
```suggestion
    GPU_OPERATOR_QUAY_BUNDLE_PUSH_SECRET=${GPU_OPERATOR_QUAY_BUNDLE_PUSH_SECRET:-"/var/run/psap-entitlement-secret/openshift-psap-openshift-ci-secret.yml"}
```

Or could also just remove the `:-` in the conditional if you prefer the if then statement

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
```diff
@@ -58,17 +58,35 @@ test_master_branch() {
...
+
+    if [ -z "${GPU_OPERATOR_QUAY_BUNDLE_IMAGE_NAME:-}" ]; then
+        GPU_OPERATOR_QUAY_BUNDLE_IMAGE_NAME=quay.io/openshift-psap/ci-artifacts
+    fi
```
dagrayvid, line 80: same as above

toolbox/gpu-operator/deploy_from_operatorhub.sh
```diff
@@ -34,16 +37,28 @@ if [[ "${1:-}" == "--help" || "${1:-}" == -h ]]; then
...
+        echo "Deploying the GPU Operator from OperatorHub using its master bundle."
+    else
+        echo "Deploying the GPU Operator from OperatorHub using bundle at $bundle."
+        ANSIBLE_OPTS="${ANSIBLE_OPTS} -e deploy_bundle_image=$bundle"
     fi
+
+elif [[ "${1:-}" == "$DEPLOY_ONLY_CLUSTERPOLICY_FLAG" ]]; then
+    ANSIBLE_OPTS="${ANSIBLE_OPTS} -e gpu_operator_deploy_from=pre-deployed"
+    echo "Not deploying the GPU Operator, only create the ClusterPolicy from the CSV."
+elif [[ "${1:-}" == "-"* ]]; then
```
dagrayvid, line 58:
```suggestion
elif [[ "${1}" == "-"* ]]; then
```
There are a few uses of this syntax in this file. Doesn't the `:-` just substitute an empty string if $1 is not set -- making it unnecessary?

omertuc: It is necessary because of `set -u`

dagrayvid: Thanks for the explanation. The previous comments about setting the variables are still valid, right?

omertuc: I think they are, yeah

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
```diff
@@ -0,0 +1,41 @@
...
+Fetch metrics from a Pod
+
+Usage: $0 <POD_PORT> <LABEL_SELECTOR> <NAMESPACE> <PROM_GREP_KEY>
+Example:
+  $0 9400 app=nvidia-dcgm-exporter gpu-operator-resources '#HELP'
+EOF
+    exit 0
+fi
+
+POD_PORT="$1"
```
omertuc, line 14: Nit: use
```bash
...=$1
shift
...=$1
shift
```
instead

It's easier to maintain - causes less diff / mistakes when modifying parameters

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
```diff
@@ -0,0 +1,41 @@
...
+
+LOCAL_PORT=9401
+
+pod_name=$(oc get pods -oname -l$LABEL_SELECTOR -n$NAMESPACE | head -1);
+if [ -z "$pod_name" ]; then
+    echo "Failed to find a pod for $LABEL_SELECTOR in namespace $NAMESPACE";
+    exit 10;
+fi;
+
+retry=5;
```
omertuc, line 27: Nit: Rename to `tries_left`

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
```diff
@@ -0,0 +1,41 @@
...
+timeout 10 oc port-forward ${pod_name} ${LOCAL_PORT}:${POD_PORT} -n $NAMESPACE &
+trap wait EXIT
+pod_output=""
+while [ "$pod_output" == "" ]; do
+    sleep 1;
+    pod_output=$(curl localhost:${LOCAL_PORT}/metrics 2>/dev/null);
```
omertuc, line 33: Since this is a proper script now, I think all those `;` are unnecessary

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
```diff
@@ -0,0 +1,41 @@
...
+LOCAL_PORT=9401
+
+pod_name=$(oc get pods -oname -l$LABEL_SELECTOR -n$NAMESPACE | head -1);
+if [ -z "$pod_name" ]; then
+    echo "Failed to find a pod for $LABEL_SELECTOR in namespace $NAMESPACE";
+    exit 10;
+fi;
+
+retry=5;
+timeout 10 oc port-forward ${pod_name} ${LOCAL_PORT}:${POD_PORT} -n $NAMESPACE &
```
omertuc, line 28: If I understand it correctly, the tunnel will get cleaned up after 10 seconds regardless of what happens, right?

kpouget:
```
oc port-forward ${pod_name} ${LOCAL_PORT}:${POD_PORT} -n $NAMESPACE &
OC_PORT_FWD_PID=$!
trap "kill -9 $OC_PORT_FWD_PID 2>/dev/null" EXIT
```
I didn't push the last changes, I've updated the code with this ^^^
so that we kill the tunnel on exit

omertuc: Wouldn't the `timeout` kill it anyway?

kpouget: I changed it because for the metrics I need to test this twice in a row (DCGM & operator), and the 2nd one failed because the port was still open! (didn't timeout yet :))

roles/gpu_operator_bundle_from_commit/files/operator_bundle_image_builder_script.yml
```diff
@@ -0,0 +1,51 @@
...
+metadata:
+  labels:
+    app: gpu-operator-ci
+  name: operator-bundle-image-builder-script
+  namespace: gpu-operator-ci
+data:
+  run_operator_bundle_image_builder.sh: |
+    #! /bin/bash
+    set -ex
+    set -o nounset
```
omertuc, line 12: Sloppy rebasing work @kpouget ! You sneaked this line in a separate commit, I want to suggest adding it only to find it in the full PR diff

kpouget: ack, I'll try to reshape the commits if it's modified back and forth in multiple commits!

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
kpouget: thanks, applied

build/root/usr/local/bin/ci_entrypoint_gpu-operator.sh
kpouget: thanks, applied

toolbox/gpu-operator/deploy_from_operatorhub.sh
kpouget: yes, it's "if undefined or empty" (I didn't know the 'or empty' part :)
```
$ var=""
$ echo ${var:-default}
default

$ echo ${not_a_var:-default}
default
```

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
kpouget: ack

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
kpouget: ack, I'm not 100% convinced for "less mistakes", but perhaps

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
kpouget: ack, it was a left-over from when I moved it from ansible task :)

roles/gpu_operator_wait_deployment/files/fetch_pod_metrics.sh
kpouget: > Typically timeout is used to protect from long running commands,

`oc port-forward` is a long method, no? it never terminates by itself

omertuc: Yes, that's why I found it surprising/confusing that `timeout` is used on `oc port-forward`, and why I asked for a comment clarifying what it's about

kpouget:
@omertuc @dagrayvid , PR has been updated following the comments, thanks for the feedback:)

/test gpu-operator-commit

kpouget:
/test gpu-operator-e2e

@omertuc can we merge this one? I added this commit because I can't find out how to properly use quay.io images ...
```
    build: test_master_branch: use test_commit instead of test_master_branch

    `test_master_branch` uses our OLM bundle + operator image
    pre/automatically built on Quay.io, but the bundle image still points
    to NVIDIA v1.7.1 version of the operator.

    I can't find a way to successfully fix the image on the fly, so this
    commit switches the `master` branch testing to a "commit" test,
    pointing to the GPU Operator master branch. This will build the
    operator image in the cluster + the bundle image with the correct
    operator image address.
```

roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_from_bundle.yml
```diff
@@ -15,18 +15,17 @@
...
+    operator-sdk run bundle
+                 -n {{ deploy_bundle_namespace }}
+                 {{ deploy_bundle_image }}
+
+- name: Fix the GPU Operator image
+  command:
+    oc set image deployment/gpu-operator
+       gpu-operator={{ psap_quay_master_operator }}
+       -n {{ deploy_bundle_namespace }}
+  when: deploy_bundle_namespace == psap_quay_master_bundle
```
omertuc, line 28: Can you explain this condition?

Why would `deploy_bundle_namespace` be equal to `psap_quay_master_bundle`

kpouget: wait, this hunk/commit is a) wrong b) not relevant anymore :|

correct condition was:
```
   when: deploy_bundle_image == psap_quay_master_bundle
```
but I couldn't get this to work properly, see https://github.com/openshift-psap/ci-artifacts/pull/203.

I'll replace this commit by the one in https://github.com/openshift-psap/ci-artifacts/pull/203 as it's still the closest to what I have in mind

kpouget:
@dagrayvid @omertuc can we merge this one for fixing the nightly testing of the GPU Operator master branch?

omertuc:
> @dagrayvid @omertuc can we merge this one for fixing the nightly testing of the GPU Operator master branch?

Still reviewing....

kpouget:
thanks @omertuc 
/approve

PR #200, opened 2021-06-22
Fix NFD image URL replacement

ArangoGutierrez:
This Patch moves from a fixed sed edit to a hard one replacing the entire line, that one sed won't fail when ever the target string is updated

### DIFF
--- a/roles/nfd_operator_deploy_custom_commit/files/operator_image_builder_script.yml
+++ b/roles/nfd_operator_deploy_custom_commit/files/operator_image_builder_script.yml
@@ -31,8 +31,8 @@ data:
 
     podman pull docker.io/library/golang:1.16.3-buster
     podman pull registry.access.redhat.com/ubi8/ubi:latest
-    sed -i 's/registry.ci.openshift.org\/ocp\/builder:rhel-8-golang-1.16-openshift-4.8/docker.io\/library\/golang:1.16.3-buster/' Dockerfile
-    sed -i 's/registry.ci.openshift.org\/ocp\/4.8:base/registry.access.redhat.com\/ubi8\/ubi/' Dockerfile
+    sed -i '2s|.*|FROM docker.io/library/golang:1.16.3-buster as builder|' Dockerfile
+    sed -i '10s|.*|FROM registry.access.redhat.com/ubi8/ubi|' Dockerfile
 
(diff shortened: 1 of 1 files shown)


roles/nfd_operator_deploy_custom_commit/files/operator_image_builder_script.yml
```diff
@@ -31,8 +31,8 @@ data:
 
     podman pull docker.io/library/golang:1.16.3-buster
     podman pull registry.access.redhat.com/ubi8/ubi:latest
-    sed -i 's/registry.ci.openshift.org\/ocp\/builder:rhel-8-golang-1.16-openshift-4.8/docker.io\/library\/golang:1.16.3-buster/' Dockerfile
-    sed -i 's/registry.ci.openshift.org\/ocp\/4.8:base/registry.access.redhat.com\/ubi8\/ubi/' Dockerfile
+    sed -i '2s/.*/FROM docker.io\/library\/golang:1.16.3-buster as builder/' Dockerfile
+    sed -i '10s/.*/FROM registry.access.redhat.com\/ubi8\/ubi/' Dockerfile
```
kpouget, line 35: @ArangoGutierrez can you change the `sed` separator? 
```sed -i '...|...|' Dockerfile```

Comment (ArangoGutierrez) [reply roles/nfd_operator_deploy_custom_commit/files/operator_image_builder_script.yml:35]: ah? I have never used sed with | in the past, you sure?

Comment (kpouget) [reply roles/nfd_operator_deploy_custom_commit/files/operator_image_builder_script.yml:35]: yes, the delimiter is free (any character?)
```sed 's#search#replace#'
sed 's@search@replace@'
```

just mind the quoting :)
```
sed s|search|replace|
# or
sed s#search#replace#
```
those won't do what you want :)

kpouget:
```
$ echo soso | sed 'sasaSag'
SoSo
```

ArangoGutierrez: I see , I have always used / in the past :) , ok will change

ArangoGutierrez: done

PR #207, opened 2021-07-02
{nfd,gpu}_operator_deploy_from_operatorhub: sed/subscriptions/subscriptions.operators.coreos.com + minor updates 

kpouget:
* 66cce04 - {nfd,gpu}_operator_deploy_from_operatorhub: sed/subscriptions/subscriptions.operators.coreos.com

`subscription` object name may be used by the KNative (in addition to
OLM):
```
$ oc get crd -A | grep subscriptions

subscriptions.messaging.knative.dev   2021-06-07T14:57:28Z
subscriptions.operators.coreos.com    2021-06-06T16:45:30Z
```

```
$ oc explain subscriptions

KIND:     Subscription

VERSION:  messaging.knative.dev/v1
```

This breaks commands expecting OLM subscription resources like
```
oc get subscription/nfd
```

---

* 737f0c5 - gpu_operator_deploy_from_operatorhub: wait for CRD instead of waiting for CR creating

---

* 413b3ea - Always store 'oc describe' into an {{ artifact_extra_logs_dir }} file

---

PR #208, opened 2021-07-02
Backport: {nfd,gpu}_operator_deploy_from_operatorhub: sed/subscriptions/subscriptions.operators.coreos.com

kpouget:
Backport of https://github.com/openshift-psap/ci-artifacts/pull/207 fix to `stable-v0.1` branch.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,19 @@
 
 (Organized release by release)
 
+Features of version 0.1.1 (July 2021)
+-------------------------------------
+
+Bug fixes
+~~~~~~~~~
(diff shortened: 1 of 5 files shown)


omertuc:
Me might want to add CI jobs to backporting branches... although annoyingly we're limited with the branch names - we should really look into that

kpouget:
I updated the commit with @jmencak's doc suggestions

> Me might want to add CI jobs to backporting branches... although annoyingly we're limited with the branch names - we should really look into that

indeed, we need to find a way to get this branch tested, but I'm not sure how ...

PR #213, opened 2021-07-08
WIP: Add NFD Operator resource status checks

courtneypacheco:
Add playbook and role to check for the NFD Operator resource statuses to ensure that the actual NFD Operator resource statuses align with the output of the NFD CRD status output.


roles/nfd_operator_check_resource_statuses/vars/main/resources.yml
```diff
@@ -0,0 +1,41 @@
+---
```
omertuc, line 1: I fear that this list is too specific to the current version of NFD, and it would be hard to maintain it and keep it in sync with changes in the NFD operator. Once something changes in the NFD operator, we would have to update this list and also have to maintain backwards compatibility, because we might be testing both older and newer versions of the NFD operator with the same ci-artifacts version.

To overcome this, I suggest that instead of hard-coding the names of the resources, maybe it would be best to just dump all resources of each kind in the `openshift-nfd` namespace.

For cluster-scoped resources such as the namespace itself, the SCC and the ClusterRoles, which are not part of the `openshift-nfd` namespace, we can hardcode for now. But this too can be solved if the nfd-operator would label its cluster-scoped resources with some label (like `"nodefeaturediscoveried.nfd.openshinshift.io-v1-resource": "true"` or something). Then we can just scan for all cluster-scoped resources with that label and dump them.

playbooks/nfd_operator_status_check.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Check the status of various NFD resources
```
omertuc, line 2: "Check the status" could be a bit ambiguous. Is the purpose to make sure the status is ok and explicitly fail if it's not ok, or is the purpose to dump and save the status of all resources for diagnostic purposes?

roles/nfd_operator_check_resource_statuses/vars/main/resources.yml
courtneypacheco: Thanks Omer! 

That makes a lot of sense -- we don't want to have difficulties in the future with maintaining the NFD Operator.

@ArangoGutierrez - what do you think about Omer's cluster-scoped resource labels suggestion? I think his suggestion would be useful to add to the NFD Operator.

roles/nfd_operator_check_resource_statuses/tasks/get_cluster_nfd_operator_resources.yml
```diff
@@ -0,0 +1,99 @@
+---
+- name: Determine the OpenShift cluster version being used
+  shell: |
+    set -o pipefail;
+    oc version -o json | jq --raw-output '.openshiftVersion' | cut -d"." -f1-2
+  register: oc_version
```
kpouget, line 6: OpenShift version is already available with `{{ openshift_release }}`
https://github.com/openshift-psap/ci-artifacts/blob/master/roles/check_deps/tasks/main.yml#L32

roles/nfd_operator_check_resource_statuses/tasks/get_cluster_nfd_operator_resources.yml
```diff
@@ -0,0 +1,99 @@
...
+      paths: '{{ tmp_cluster_nfd_operator }}/build/assets/master'
+      patterns: '*.yaml,*.yml'
+    register: all_nfd_operator_master_manifests
+
+  # Label the manifests by resource type. Note that some manifests *may* have multple
+  # resources defined in them, hence a bunch of 'if-then' statements and not 'if-elif'
+  # statements.
+  - name: For all the worker and master manifests, label the manifest by resource type
+    shell: |
+      set -o pipefail;
```
kpouget, line 39: maybe you can you move this into a `template` file?

roles/nfd_operator_check_resource_statuses/tasks/get_cluster_nfd_operator_resources.yml
```diff
@@ -0,0 +1,99 @@
...
+    failed_when: false
+
+  - name: Save manifest names
+    set_fact:
+      nfd_operator_manifests: '{{ all_nfd_operator_manifests }}'
+
+  when: >
+    oc_version.stdout_lines[0] == "4.8" or
+    oc_version.stdout_lines[0] == "4.9" or
+    oc_version.stdout_lines[0] == "4.10"
```
kpouget, line 99: maybe you can move the `when` before the `block:` to ease the reading? (and add a `name:`, which is made mandatory by `ansible-lint`, I wonder why it wasn't picked up ...)

roles/nfd_operator_check_resource_statuses/tasks/service_init.yml
```diff
@@ -0,0 +1,44 @@
...
+  with_items: '{{ all_nfd_operator_manifests.results | selectattr(search_key,"contains",search_val) | list }}'
+  loop_control:
+    loop_var: service_manifest
+    label: '{{ service_manifest.stdout_lines }}'
+  vars:
+    search_key: 'stdout'
+    search_val: '{{ is_service }}'
+
+# Attempt to retreive the Service from the cluster to see if they exist
+- name: Attempt to retreive the NFD Operator Service
```
kpouget, line 25: retrieve

roles/nfd_operator_check_resource_statuses/vars/main/resources.yml
kpouget: > To overcome this, I suggest that instead of hard-coding the names of the resources, maybe it would be best to just iterate on all resources of each kind in the openshift-nfd namespace.

@courtneypacheco , I agree with @omertuc that you should try to avoid hard-coding the names of the resources
```
 # If no $resource were found at all, then something is wrong.
 # Attempt to retreive the $resource from the cluster to see if they exist
 # If any of the NFD Operator $resource do not exist, then we cannot continue
```

most of the files follow this pattern, I think there is a way for Ansible to move the resource type to a variable, something like
```

- name: Attempt to retreive the NFD Operator {{ resource }}
  shell: |
    set -o pipefail;
    oc get {{ resource }} '{{ resource_name.stdout }}' -n '{{ nfd_operator_namespace }}' -oname | grep 'error'
```
and maybe at the end of the generic file, `include` the resource-specific  file for further testing, if it exists

roles/nfd_operator_check_resource_statuses/tasks/clusterrole_init.yml
```diff
@@ -0,0 +1,44 @@
...
+  when: all_nfd_operator_manifests.results | selectattr(search_key,'contains',search_val) | list | count == 0
+  vars:
+    search_key: 'stdout'
+    search_val: '{{ is_clusterrole }}'
+
+# Get Role names
+- name: Get the NFD Operator's ClusterRole names from the manifests
+  shell: |
+    set -o pipefail;
+    cat '{{ clusterrole_manifest.manifest.path }}' | shyaml get-value metadata.name
```
kpouget, line 14: I'm not sure what is `shyaml`, but it's not yet part of the image for sure
we usually use `cat $file.yml | yq | jq ....` to convert yaml to json (`yq`), and then select from json (`jq`)

roles/nfd_operator_check_resource_statuses/tasks/clusterrole_init.yml
```diff
@@ -0,0 +1,44 @@
...
+  delay: '{{ nfd_operator_resource_delay }}'
+  register: nfd_operator_clusterroles_do_not_exist
+  failed_when: false
+  with_items: '{{ nfd_operator_clusterrole_names.results }}'
+  loop_control:
+    loop_var: clusterrole_name
+    label: '{{ clusterrole_name.stdout }}'
+
+- name: If any of the NFD Operator ClusterRoles do not exist, then we cannot continue
+  fail: msg="All NFD Operator ClusterRoles must be initialized before running this role"
```
kpouget, line 39: is there a way (in the CI code) to wait for NFD-Operator initialization?

playbooks/nfd_operator_status_check.yml
```diff
@@ -0,0 +1,7 @@
+---
```
kpouget, line 1: I don't see any `toolbox` script nor [`ci_entrypoint_nfd-operator.sh`](https://github.com/openshift-psap/ci-artifacts/blob/master/build/root/usr/local/bin/ci_entrypoint_nfd-operator.sh) call to this playbook (the entrypoint should call the toolbox, and the toolbox launches the playbook)

before the review can be finalized we'll need to have the code running as a PR test (with `/test nfd-operator-e2e-master`)

kpouget (changes requested):

ArangoGutierrez:
image

roles/nfd_operator_check_resource_statuses/vars/main/resources.yml
ArangoGutierrez: Agree with @omertuc , `oc get all` on the `openshift-nfd` NS should be enough resources to chaos test NFD and then see if the operator is checking the status of the given objects

kpouget:
@ArangoGutierrez I'll close this PR, it's outdated now and Courtney's gone

/close

> image

the test ^^^ doesn't execute the code of the PR, see this comment:

> I don't see any toolbox script nor `ci_entrypoint_nfd-operator.sh` call to this playbook (the entrypoint should call the toolbox, and the toolbox launches the playbook)

can always be reopened or reworked in a new PR

PR #217, opened 2021-07-14
Append operators.coreos.com to tasks deploying GPU operator from catalog

ArangoGutierrez:
PR https://github.com/openshift-psap/ci-artifacts/pull/207 started to fix this, this patch just add two more lines to the fix
Append operators.coreos.com to tasks deploying GPU operator from catalog


omertuc:
It should be `subscriptions.operators.coreos.com` rather than `Subscription.oper...`

ArangoGutierrez:
> It should be `subscriptions.operators.coreos.com` rather than `Subscription.oper...`
> 
> image

Thanks Omer! , also will fix the commit to be upper case :)

omertuc:
Also `s/Subscriptions/subscriptions/g`

PR #220, opened 2021-07-22
entitlement_test_wait_deployment: keep extra repo enabled for listing available kernels

kpouget:


kpouget:
> /lgtm

thanks,
let's merge this, it will make `entitlement_test_wait_deployment` more consistent with what the GPU Operator does, and more generally what has to be done to fetch the node's kernel packages.

I wanted to test on OCP 4.7, but default is now 4.8 and `/test 4.7-gpu-operator-e2e` wasn't created. I'll create a PR on openshift/release to test it

/approve

PR #230, opened 2021-07-29
Update docs after python re-write

ArangoGutierrez:
Some missing notes on the docs :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -27,11 +27,11 @@ Dependencies
 
 Requirements:
 
-- ``ansible`` (`>= 2.9.5`), ``yq``, ``jq``
+- See requirements.txt for reference
 
 .. code-block:: shell
 
-    pip3 install yq ansible==2.9.*
+    pip3 install -r requirements.txt
     dnf install jq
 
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -27,11 +27,11 @@ Dependencies
 
 Requirements:
 
-- ``ansible`` (`>= 2.9.5`), ``yq``, ``jq``
+- ``ansible`` (`>= 2.9.5`), ``yq``, ``jq``, ``fire``
```
omertuc, line 30: Good catch.

Maybe let's just point users to the `requirements.txt` file, rather than maintaining these too lists

README.rst
```diff
@@ -27,11 +27,11 @@ Dependencies
...
 
 .. code-block:: shell
 
-    pip3 install yq ansible==2.9.*
+    pip3 install yq ansible==2.9.* fire
```
omertuc, line 34: Here we can change to `pip3 install -r requirements.txt`

ArangoGutierrez: good idea!

PR #234, opened 2021-08-06
toolbox: cluster: add --base_machineset to set_scale

kpouget:


docs/toolbox/cluster.rst
```diff
@@ -26,3 +26,17 @@ Cluster Scale
...
     # to 0 to achive that.
     ./run_toolbox.py cluster set_scale g4dn.xlarge 5 --force
+
+ .. code-block:: shell
+
+    # list the machinesets of the cluster
+    $ oc get machinesets -n openshift-machine-api
+
+    NAME                                      DESIRED   CURRENT   READY   AVAILABLE   AGE
+    <cluster>-8p9vm-worker-eu-central-1a      1         1         1       1           57m
```
omertuc, line 36: Since this is already an example, might as well put the actual cluster name in there, I think it makes things less abstract / confusing

toolbox/cluster.py
```diff
@@ -6,7 +6,7 @@ class Cluster:
     Commands relating to cluster scaling, upgrading and environment capture
     """
     @staticmethod
-    def set_scale(instance_type, scale, force=False):
+    def set_scale(instance_type, scale, base_machineset="", force=False):
```
omertuc, line 9: Might want to fix the docstring - it doesn't explain anymore what happens when a user specifies a `base_machineset`

toolbox/cluster.py
```diff
@@ -24,14 +24,16 @@ def set_scale(instance_type, scale, force=False):
...
         """
         opts = {
                 "machineset_instance_type": instance_type,
                 "scale": scale,
+                "base_machineset": base_machineset,
             }
 
         if force:
-            opts = {**opts, "force_scale": "true"}
+            opts["force_scale"] = "true"
```
omertuc, line 36: Sometimes I do things in ridiculous ways when I forget there are absolutely trivial ways to do them :sweat_smile:

docs/toolbox/cluster.rst
kpouget: changed `<cluster>` to `playground`

toolbox/cluster.py
kpouget: why not? there is this below (I'll removed the `s` of `defaults`)
```
             base_machineset: Name of a machineset to use to derive the new one. Defaults: pickup the first machineset found in `oc get machinesets -n openshift-machine-api`.
```

toolbox/cluster.py
kpouget: 🙃 😉

toolbox/cluster.py
omertuc: The entire description about the parameters docs doesn't take this parameter into account, and it's unclear what would happen when it's specified

kpouget: updated

kpouget:
ping @omertuc

kpouget:
thanks Omer,

/approve

kpouget:
hum, shoot, this PR wasn't tested, and it lead to a CI failure :/
Fixed in https://github.com/openshift-psap/ci-artifacts/pull/243, thanks @omertuc

PR #240, opened 2021-08-10
Allow no action on the custom run command

omertuc:


build/root/usr/local/bin/run
```diff
@@ -99,10 +99,10 @@ trap "postchecks ERR" ERR
 target="${1:-}"
 shift
 action="${1:-}"
-shift
+shift || true
 
 echo
-echo "Test target: '$target'; test action: '$action'"
+echo "Test target: '$target'; test action: '${action:-}'"
```
kpouget, line 105: I think this change isn't needed, as `action` is always set by ` action="${1:-}"`

PR #243, opened 2021-08-16
Better handling of empty base_machineset

omertuc:
Before this change, not specifying the base machineset would cause an
error because the ansible code treated an empty "" string differently
from an unset variable altogether.

This change makes it so the toolbox script leaves the base_machineset
variable empty, rather than have it be an empty string


jmencak:
Thank you for the PR, Omer.

/lgtm

ArangoGutierrez (approved):
/lgtm

PR #244, opened 2021-08-24
New role gpu_operator_get_csv_version and other improvements

kpouget:
* `entitlement_test_wait_deployment: do not check openssl x509 -enddate if entitlement not deployed`

---

* `gpu_operator_deploy_from_operatorhub: deploy_from_bundle: give 5min to deploy instead of 2min`

---

* `gpu_operator_get_csv_version: new role to simplify version-conditional execution of ansible tests`

---

* `toolbox: _common.py: allow overriding ARTIFACT_EXTRA_LOGS_DIR`

---

* `toolbox/gpu-operator/must-gather.sh: capture GPU Operator CSV version`

---

* `CI: GPU Operator: store in ${ARTIFACT_DIR} the operator/cluster/ci-artifacts version`

For display in ci-dashboard

---

* `nfd_operator_undeploy_from_operatorhub: undeploy from the right namespace`

---

* `gpu_operator_run_gpu-burn: extend wait time to 10min`

---

* `CI: ci_entrypoint_gpu-operator.sh: mute 'entitlement undeploy' finalizer`

---


toolbox/_common.py
```diff
@@ -59,11 +59,12 @@ def run_ansible_playbook(playbook_name, opts: dict = dict()):
     artifact_dir = Path(os.environ["ARTIFACT_DIR"])
     artifact_dir.mkdir(parents=True, exist_ok=True)
 
-    previous_extra_count = len(list(artifact_dir.glob("*__*")))
-    os.environ["ARTIFACT_EXTRA_LOGS_DIR"] = str(
+    if os.environ.get("ARTIFACT_EXTRA_LOGS_DIR") is None:
```
omertuc, line 62: The original `_common.sh` file is still in use by some scripts so you might want to update it as well

roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_clusterpolicy.yml
```diff
@@ -48,26 +48,12 @@
...
-  command: oc get clusterpolicies.nvidia.com -ojsonpath='{range .items[*]}{.spec.validator.version}{end}'
-  register: gpu_operator_atleast_v170
-  failed_when: false
+- name: Get the GPU Operation version
+  include_role:
+    name: gpu_operator_get_csv_version
 
 - name: Create the namespace for the GPU Operator (only < v1.7.0)
-  when: not gpu_operator_atleast_v170.stdout
+  when: gpu_operator_version < "1.7.0"
```
omertuc, line 56: Does string comparison work like this? Is it alphabetical?

omertuc: Wouldn't it break with 2-digit version components? e.g. 1.10.2

omertuc:
Other than comments looks good

roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_clusterpolicy.yml
kpouget: yes for the part "works like this", I double checked,
but indeed it will break > 1.9 ;/

omertuc: https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#comparing-versions

kpouget: great, thanks, I'll update the PR

toolbox/_common.py
kpouget: ack, fixed

roles/gpu_operator_deploy_from_operatorhub/tasks/deploy_clusterpolicy.yml
kpouget: fixed

kpouget:
thanks @omer,

/approve

kpouget:
/approve cancel

there is a race condition in `gpu-operator-e2e` caused by `013__entitlement__undeploy` (in `presubmit-operatorhub`), which deletes the entitlement, but doesn't wait for all the nodes to be rebooted. After that, in `presubmit-master`, `001__entitlement__test_cluster` passes, but the driver deployment fails because the entitlement is gone ...

kpouget:
> there is a race condition in `gpu-operator-e2e` caused by `013__entitlement__undeploy` (in `presubmit-operatorhub`), which deletes the entitlement, but doesn't wait for all the nodes to be rebooted. After that, in `presubmit-master`, `001__entitlement__test_cluster` passes, but the driver deployment fails because the entitlement is gone ...

I added a commit to wait for the proper undeployment of the `MachineConfig` before running `presubmit-master` test

kpouget:
@omertuc any clue why `| jq` works on my system, but not in the CI (it needs `| jq .`)
version mismatch?

PR #251, opened 2021-09-07
.github/workflows/revert-testing.yml: new post-push action to remove tmp test commits

kpouget:
This PUSH Github action will revert the last commit of a PR being
merged if its message starts with '[TEST TMP]'.

This allows running custom PR tests in OpenShift CI, without affecting
the routes taken by the nightly CI.

### DIFF
--- a/.github/workflows/revert-testing.yml
+++ b/.github/workflows/revert-testing.yml
@@ -1,0 +1,40 @@
+name: 'Revert "[TEST TMP]" commit'
+on:
+  push:
+    branches: [master]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: 'Revert "[TEST TMP]" commit'
+        run: |
+          set -euxo pipefail;
+
+          git config --global user.email "github.action@ci-artifacts.psap"
+          git config --global user.name "Github Action"
+
+          COMMIT_MSG=$(git log --pretty='format:%Creset%s' --no-merges -1)
+          COMMIT_AUTHOR=$(git show -s --format='%an <%ae>' HEAD)
+
+          if ! echo "$COMMIT_MSG" | grep '^\[TEST TMP\]' -q; then
+            echo "Not a temporary test commit, do nothing."
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/revert-testing.yml
```diff
@@ -0,0 +1,40 @@
...
+          git config --global user.name "Github Action"
+
+          COMMIT_MSG=$(git log --pretty='format:%Creset%s' --no-merges -1)
+          COMMIT_AUTHOR=$(git show -s --format='%an <%ae>' HEAD)
+
+          if ! echo "$COMMIT_MSG" | grep '^\[TEST TMP\]' -q; then
+            echo "Not a temporary test commit, do nothing."
+          fi
+
+          echo "Temporary test commit detected, reverting $COMMIT_MSG ..."
```
omertuc, line 36: Any reason not to simply `git reset --hard HEAD~`?

kpouget: yes, `reset` removes the commit, while `revert` keeps it (both of them) in the history

I like the idea of keeping the trace of how the PR was tested as part of the history

but result is identical indeed

.github/workflows/revert-testing.yml
```diff
@@ -0,0 +1,40 @@
...
+          git show --stat HEAD
+          git revert HEAD
```
omertuc, line 38: Don't you need `--no-edit` for immediate revert?

.github/workflows/revert-testing.yml
omertuc: Yeah keeping it definitely makes sense

.github/workflows/revert-testing.yml
kpouget: good point, I guess that `git` checks if there is no `tty` and skips this part, as the action did worked

```
       -e, --edit
           With this option, git revert will let you edit the commit message
           prior to committing the revert. This is the default if you run the
           command from a terminal.
```

omertuc: > good point, I guess that `git` checks if there is no `tty` and skips this part, as the action did worked
> 
> ```
>        -e, --edit
>            With this option, git revert will let you edit the commit message
>            prior to committing the revert. This is the default if you run the
>            command from a terminal.
> ```

I tried piping it into `cat` locally and it didn't detect no tty. Odd

kpouget:
```
true | git revert HEAD | cat
```

omertuc: `echo '' | git revert` works too, so it only checks if `stdin` is a tty or not... Never seen this behavior on any other program

kpouget:
@omertuc good for merging I think, no comment on slack == no objection I assume :)

omertuc:
> @omertuc good for merging I think, no comment on slack == no objection I assume :)

Nah that's just my forgetfulness 😅

kpouget:
thanks @omertuc  ;-)

PR #253, opened 2021-09-08
.github/workflows/revert-testing.yml: exit if revert not required

kpouget:
- `exit` if revert not required :|
- and add a `[TEST TMP]` commit to validate that the action works as expected
- and add `origin master` to `git push` to specify where to push
- and make `git amend` `--quiet`


kpouget:
> - and add origin master to git push to specify where to push

but this won't be enough to pass over the protected branch hook barrier :(

>  ! [remote rejected] master -> master (protected branch hook declined)
> error: failed to push some refs to 'https://github.com/openshift-psap/ci-artifacts'

kpouget:
@omertuc can we merge this, to avoid trying to revert every top-most commit ? 
> exit if revert not required :|
🙃

PR #272, opened 2021-10-20
Remove fuse NTO profile

kpouget:
Fuse module is now loaded by default by RHCOS, and OSD doesn't allow
the node tagging we do to load fuse only on a single node.


kpouget:
flake

/test 4.6-gpu-operator-e2e

PR #292, opened 2021-12-01
Adding Image Build/Push Role + Recipe for MLPerf SSD Image

Maxusmusti:
Adding role for building images and pushing to quay, and a recipe for building/pushing mlperf ssd image

NOTE: `test-path: recipes publish_mlperf_ssd_image`


toolbox/gpu_operator.py
```diff
@@ -211,6 +211,46 @@ def to_y(_s):
 
         return PlaybookRun("gpu_operator_bundle_from_commit", opts)
 
+    @staticmethod
+    def build_push_image(
```
omertuc, line 215: What does this have to do with the GPU operator? It seems pretty generic

Maxusmusti: It doesn't specifically. I only had it included with GPU operator as it's intended use-case was with it. Unsure if it will need anything GPU operator specific though, and if not I will likely abstract it. Will ask @kpouget about it in our meeting tomorrow morning.

kpouget:
thanks for the PR @Maxusmusti, 

as @omertuc mentioned, this isn't related to the GPU Operator [CI], so let's move it out of the `gpu_operator`, but I'm not sure where to have it. `cluster`? but it's not cluster related ... `utils`?

I was going to add that `podman` isn't available in the [image](https://github.com/openshift-psap/ci-artifacts/blob/master/build/Dockerfile) (this still stands), but it seems to be available without entitlement in UBI, so seems we'll be good with `podman` package installed in the Dockerfile. IIRC, it required RHEL entitlement until recently, but I've just installed it in a non-entitled cluster

roles/gpu_operator_build_push_image/files/build-push-image.sh
```diff
@@ -0,0 +1,59 @@
...
+        d) dockerfile=${OPTARG};;
+        p) path=${OPTARG};;
+        q) quay="quay.io/${OPTARG}";;
+        a) authfile=${OPTARG};;
+        *) exit 1 ;;
+    esac
+done
+
+if [[ -z $repo && -z $dockerfile ]] ; then
+    printf -- "Either a git repo (-g) or Dockerfile (-d) is required\n"
```
kpouget, line 33: why `printf` instead of `echo`?

roles/gpu_operator_build_push_image/files/build-push-image.sh
```diff
@@ -0,0 +1,59 @@
...
+    exit 1
+elif [[ -z $name || -z $tag ]] ; then
+    printf -- "Image name (-n) and tag (-t) required\n"
+    exit 1
+fi
+
+if [[ -n $dockerfile ]] ; then
+    podman build -t $name:tag -f $dockerfile
+elif [[ -n $repo ]] ; then
+    git clone $repo
```
kpouget, line 49: I would suggest:
```
git clone $repo repo
cd repo
...
```
and change
```
 path="training_results_v0.7/NVIDIA/benchmarks/ssd/implementations/pytorch"
#to  
 path="/NVIDIA/benchmarks/ssd/implementations/pytorch"
```
(I think quay.io requires the initial `/`, like an absolute path, I don't know if it's a common convention or not)

roles/gpu_operator_build_push_image/files/build-push-image.sh
```diff
@@ -0,0 +1,59 @@
...
+    podman build -t $name:tag -f $path
+fi
+
+if [[ -n ${quay:8} ]] ; then
```
kpouget, line 53: not sure what is this `8`?

what about this?
```
if [[ "${quay:-}" ]]; ...
```

testing/recipes/publish_mlperf_ssd_image.sh
```diff
@@ -0,0 +1,15 @@
...
+repo="https://github.com/openshift-psap/training_results_v0.7.git"
+path="training_results_v0.7/NVIDIA/benchmarks/ssd/implementations/pytorch"
+quay="openshift-psap/nvidiadl-ssd-training-benchmark"
+auth="UNKNOWN"
+
+./run_toolbox.py gpu_operator build_push_image $local_name $tag \
+    --git_repo=$repo \
+    --git_path=$path \
+    --quay_org_repo=$quay \
+    --auth_file=$auth
```
kpouget, line 15: our quay push secret is there:

```/var/run/psap-entitlement-secret/openshift-psap-openshift-ci-secret.yml```

File: roles/gpu_operator_build_push_image/files/build-push-image.sh (hunk 3 above)
Comment (Maxusmusti) [reply roles/gpu_operator_build_push_image/files/build-push-image.sh:49]: Note: also use shallow

**Comment by kpouget** [comment]
```
 FATAL: Unknown entrypoint 'recipe' | publish_mlperf_ssd_image 
```
in `test-path: recipe publish_mlperf_ssd_image` it should be `recipes`, as in the dir name ;-)

File: toolbox/gpu_operator.py (hunk 1 above)
Comment (kpouget) [inline toolbox/gpu_operator.py:215]: need to remove this block now

File: roles/utils_build_push_image/files/build-push-image.sh [hunk 6]
```diff
@@ -0,0 +1,60 @@
...
+    exit 1
+elif [[ -z $name || -z $tag ]] ; then
+    echo -- "Image name (-n) and tag (-t) required"
+    exit 1
+fi
+
+if [[ -n $dockerfile ]] ; then
+    podman build -t $name:tag -f $dockerfile
+elif [[ -n $repo ]] ; then
+    git clone --depth 1 $repo repo
```
kpouget, line 49: you need to provide a `ref` here, eg `--branch $ref`, so that you can allow cloning another branch/tag/commit than the default `master` branch

and can you add a flag to make it shallow: `--depth 1` I think

roles/utils_build_push_image/files/build-push-image.sh
```diff
@@ -0,0 +1,66 @@
+#! /bin/bash
+
```
kpouget, line 2: you'll need that here:
```
set -o errexit
set -o pipefail
set -o nounset
```
https://medium.com/factualopinions/consider-starting-all-your-bash-scripts-with-these-options-74fbec0cbb83

I expect `podman build` to fail because you're currently using the (implicit) `master` branch of the repo

could you cherry pick this commit: https://github.com/kpouget/training_results_v0.7/commit/168f1fae6d91082d775ab698b1cb49cc3bc4d27a in a new branch, and give it a meaningful name (`fix/build-error`?), so that we can keep it there in the long term?

testing/recipes/publish_mlperf_ssd_image.sh
```diff
@@ -0,0 +1,15 @@
...
+
+local_name="ssd"
+tag="latest"
+
+repo="https://github.com/openshift-psap/training_results_v0.7.git"
+path="/NVIDIA/benchmarks/ssd/implementations/pytorch"
+quay="openshift-psap/nvidiadl-ssd-training-benchmark"
+auth="/var/run/psap-entitlement-secret/openshift-psap-openshift-ci-secret.yml"
+
+./run_toolbox.py gpu_operator build_push_image $local_name $tag \
```
kpouget: you'll need to add a `branch` name here as well, see the other comment

kpouget:
```
<stderr> time="2021-12-08T14:51:35Z" level=warning msg="\"/\" is not a shared mount, this could cause issues or missing mounts with rootless containers"
<stderr> time="2021-12-08T14:51:35Z" level=error msg="cannot find UID/GID for user psap-ci-runner: No subuid ranges found for user \"psap-ci-runner\" in /etc/subuid - check rootless mode in man pages."
<stderr> time="2021-12-08T14:51:35Z" level=warning msg="using rootless single mapping into the namespace. This might break some images. Check /etc/subuid and /etc/subgid for adding sub*ids"
<stderr> Error: context must be a directory: "/NVIDIA/benchmarks/ssd/implementations"
----- FAILED ----
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/292/pull-ci-openshift-psap-ci-artifacts-master-test-commit/1468592601027317760/artifacts/test-commit/test-commit/artifacts/recipes_publish_mlperf_ssd_image/test-log.txt
doesn't smell good :'(

could you add a `set -x` at the beginning of the script to better understand which step fails?

roles/utils_build_push_image/files/build-push-image.sh
```diff
@@ -0,0 +1,71 @@
...
+set -o errexit
+set -o pipefail
+set -o nounset
+
```
kpouget, line 6: can you add a `set -x` here, to make the error more obvious?

roles/utils_build_push_image/files/build-push-image.sh
```diff
@@ -0,0 +1,71 @@
...
+        git clone --depth 1 $repo --branch $branch repo
+    else
+        git clone --depth 1 $repo repo
+    fi
+    cd repo
+    full="$(pwd)$path"
+    podman build -t $name:tag -f $full
+fi
+
+if [[ -n ${quay:8} ]] ; then
```
kpouget, line 65: could you turn this (at the beginning of the file):
```
quay="quay.io/${OPTARG}";;
```
into this
```
quay="${OPTARG}";;
```
so that this `${quay:8}` can be simplified into `${quay}`?

my goal with that is to make it obvious what the test if doing. Here, to understand what `${quay:8}` does, you need to look up the assignment, notice the `quay.io/` hardcoded prefix, (and maybe google bash `${var:n}` syntax)

maybe below, define  `dest=quay.io/$quay:$tag`, and/or change the var name `quay` into `image` or something like that

build/Dockerfile
```diff
@@ -58,6 +59,11 @@ ENV USER_NAME=psap-ci-runner \
 
 COPY . ${HOME}/
 
+RUN echo "psap-ci-runner:200000:1000" >> /etc/subuid \
+	&& echo "psap-ci-runner:200000:1000" >> /etc/subgid \
+	&& chmod 777 /usr/bin/newgidmap \
+	&& chmod 777 /usr/bin/newuidmap
```
kpouget, line 65: I don't follow what's happening here,

```
time="2022-01-02T01:29:52Z" level=warning msg="\"/\" is not a shared mount, this could cause issues or missing mounts with rootless containers"
Error: cannot setup namespace using newuidmap: fork/exec /usr/bin/newuidmap: operation not permitted
```

the files exist, otherwise `chmod` would crash, but the `chmod 777` is to get rid of the `operation not permitted` error?
I don't think it's a file permission error, otherwise it would fail with a message like this:
```
./mig-manager.log
zsh: permission denied: ./mig-manager.log
```
```
EPERM 1 Operation not permitted
# vs
EACCES 13 Permission denied
```

so must be some cgroup things being denied ...
Did you manage to reproduce the error in a local cluster?

Prow (OpenShift) sets some UID restrictions, see [this](https://stackoverflow.com/questions/68430367/openshift-is-running-all-pods-with-the-first-uid-from-the-range) and maybe [that](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids), I had issues around this topic during the early setup of the CI, but I'm not sure how to create Pods with such restrictions ... (let me know if you find out)

Maxusmusti:
\test test-commit

build/Dockerfile
```diff
@@ -58,6 +61,13 @@ ENV USER_NAME=psap-ci-runner \
 
 COPY . ${HOME}/
 
+RUN echo "psap-ci-runner:231072:65536" >> /etc/subuid \
+	&& echo "psap-ci-runner:231072:65536" >> /etc/subgid \
+	&& chmod 7777 /usr/bin/newgidmap \
+	&& chmod 7777 /usr/bin/newuidmap \
+	&& echo “user.max_user_namespaces=28633” > /etc/sysctl.d/userns.conf \
+	&& sysctl -p /etc/sysctl.d/userns.conf
```
kpouget, line 69: hey @Maxusmusti, this line sounds like you're confused about how containers works, the image vs the runtime, let me know if you want to discuss that. This command (`sysctl -p`) can be executed from the script entrypoint, but not sure if it's allowed or not to change this setting from inside a container

this commit as well, as grub is never ever involved as part of a container ...
https://github.com/openshift-psap/ci-artifacts/pull/292/files/46a3a783845865848c8eacee0cc7e33c313ffc67..90d7b1e3d926e7fc295b34f4847819383d9a3734

all of these attempts seem to say there won't be any solution to run Podman in the Prow test Pod ....

roles/utils_build_push_image/files/create-build-config.sh
```diff
@@ -0,0 +1,76 @@
+#! /bin/bash
```
kpouget, line 1: could you get rid of this file, and move everything in the ansible? templating will be more robust this way

roles/utils_build_push_image/vars/main/resources.yml
```diff
@@ -0,0 +1,4 @@
+---
+# Script for building and pushing images
+utils_create_build_config_script: roles/utils_build_push_image/files/create-build-config.sh
+build_config_output_dir: roles/utils_build_push_image/files
```
kpouget, line 4: output dir should always be `{{ artifact_extra_logs_dir }}`, except when dealing with secret information (you're not, at this point)

toolbox/utils.py
```diff
@@ -0,0 +1,52 @@
...
+            git_repo - Git repo containing Dockerfile if used as source
+            git_path - Path to Dockerfile in git repo
+            branch - Branch of repo to clone
+            quay_org_repo - Org/Repo in quay to push to
+            auth_file - Auth file for quay
+        """
+
+        if not image_tag:
+            image_tag = secrets.token_hex(4)
+
```
kpouget, line 40: you can do all the parameter checks at this level
```
if [[ -z $repo && -z $dockerfile ]] ; then
    echo "Either a git repo (-g) or Dockerfile (-d) is required"
    exit 1
elif [[ -n $repo && -n $dockerfile ]] ; then
    echo "Cannot have both -g and -d"
    exit 1
elif [[ -n $repo && -z $path ]] ; then
    echo -- "Must supply path to Dockerfile within git repo (-p)"
    exit 1
elif [[ -z $name || -z $tag ]] ; then
    echo -- "Image name (-n) and tag (-t) required"
    exit 1
elif [[ -z $outdir ]] ; then
    echo -- "Build config output dir (-o) required"
    exit 1
fi
```

testing/recipes/publish_mlperf_ssd_image.sh
```diff
@@ -0,0 +1,23 @@
...
+
+set -o errexit
+set -o pipefail
+set -o nounset
+
+local_name="ssd"
+tag="nvidiadl-ssd-training-benchmark"
+
+repo="https://github.com/openshift-psap/training_results_v0.7.git"
+path="NVIDIA/benchmarks/ssd/implementations/pytorch"
```
kpouget, line 11: can you add a comment saying that we use a custom commit to avoid an image build bug?

I'll repush my commit https://github.com/kpouget/training_results_v0.7/commit/168f1fae6d91082d775ab698b1cb49cc3bc4d27a in this branch with the right git message

kpouget (changes requested):
thanks @Maxusmusti, looks good :)
I just added comments regarding the naming, nothing serious!

roles/utils_build_push_image/tasks/main.yml
```diff
@@ -0,0 +1,200 @@
...
+  template:
+    src: "{{ utils_repo_build_config }}"
+    dest: "{{ artifact_extra_logs_dir }}/config.yml"
+    mode: 0400
+  when: git_repo != ""
+
+- name: Grab local Dockerfile content
+  command: sed 's/^/      /' {{ docker_path }}
+  register: dockerfile
+  when: docker_path != ""
```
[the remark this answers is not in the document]
Maxusmusti: The sed command adds the indent in front of every line in the file, whereas putting it in the yaml will only indent the first line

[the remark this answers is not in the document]
kpouget: hum, I missed that 🙃 
so, something [like](https://stackoverflow.com/a/31856334/341106) this then:
```
  source:
    dockerfile: |
{{ dockerfile.stdout | indent(4, True) }}
```

roles/utils_build_push_image/tasks/main.yml
```diff
@@ -0,0 +1,200 @@
...
+  fail: msg="Bailing out. This play requires 'branch'"
+  when: branch is undefined
+
+- name: Check that artifact_extra_logs_dir is defined
+  fail: msg="Bailing out. This play requires 'artifact_extra_logs_dir'"
+  when: artifact_extra_logs_dir is undefined
+
+- name: Check that memory is defined
+  fail: msg="Bailing out. This play requires 'memory'"
+  when: memory is undefined
```
kpouget, line 19: will this work with no `memory` is passed to the CLI?

and could you add a file `config/main.yml` like this [one](https://github.com/openshift-psap/ci-artifacts/blob/master/roles/gpu_operator_bundle_from_commit/defaults/main/config.yml) to list the parameters expected by the role?

kpouget (changes requested):
thanks @Maxusmusti, looks quite good, I pointed out a few details to improve, nothing serious

kpouget:
thanks @Maxusmusti looks good,
last thing, `Run ansible-lint on the playbooks and roles / build (pull_request) ` --> this Action is broken, could you run this manually and try to fix whatever it reports?
```
ansible-lint -c ./config/ansible-lint.yml roles playbooks
```
(we're trying to follow Ansible best/recommended practises, one of them is "every task should have a name" ... including `block/always/rescue` 🙃 Just put a description of the block there)

Maxusmusti:
Made all necessary changes to remove all warnings/suggestions from ansible-lint and comply with best practices

kpouget:
thanks for the effort @Maxusmusti, LGTM
/lgtm
/approve
/override ci/prow/gpu-operator-e2e

kpouget:
just some ideas maybe for a follow up update of this role:
```
image_buildconfig.yml
helper-image-push-pod.yml
helper.yml
```
[these files](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/292/pull-ci-openshift-psap-ci-artifacts-master-test-commit/1486596013656379392/artifacts/test-commit/test-commit/artifacts/recipes_publish_mlperf_ssd_image/000__utils__build_push_image/) are the one doing the job, they could be named with `000n_<name>.yaml` to make it easier to reproduce outside of the CI (see [there](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-rh-ecosystem-edge-ci-artifacts-master-4.9-gpu-operator-e2e-master/1486474237433614336/artifacts/gpu-operator-e2e-master/nightly/artifacts/005__nfd_operator__deploy_from_operatorhub/) for an example)
and also for reproducibility, remove `namespace: utils-ci` from the yaml template, add create with `oc create -f ... -n utils-ci`

PR #299, opened 2021-12-07
testing: test-commit.sh: allow testing any commit of a PR

kpouget:
`test-commit.sh` added the ability to trigger a test not hard-coded into `openshift/release`, like `/test gpu-operator-e2e` would do.

Instead, currently (before this PR), we can add to the top most commit of a PR:
```
test-path: recipes test "Hello World"
```
and trigger it with
```
/test test-commit
```
which will execute `run recipes test "Hello World"` in the image.

This helps running custom one-off tests like running [this script](https://github.com/openshift-psap/ci-artifacts/blob/master/testing/recipes/run_cocodataset_nvidiadl_ssd.sh):
```
test-path: recipes run_cocodataset_nvidiadl_ssd
```
---
One restriction of the current code is that the `test-path` must be in the top most commit (think `git show HEAD`), 
and that makes `/test test-commit` fail when you add another commit on top of the one defining the `test-path`.

This PR allows [`test-commit.sh`](https://github.com/openshift-psap/ci-artifacts/blob/master/testing/test-commit.sh) to look at all the commit messages of the PR.

### DIFF
--- a/testing/test-commit.sh
+++ b/testing/test-commit.sh
@@ -26,17 +26,19 @@ commit="HEAD"
 parent=$(git log --pretty=%P -n 1 $commit)
 
 if grep -q " " <<< "$parent"; then
-    commit=$(cut -d" " -f2 <<< "$parent")
-    echo "HEAD is a merge commit. Taking the 2nd parent from $parent"
+    commit=$(sed 's/ /../g' <<< "$parent")
+    echo "HEAD is a merge commit. Taking all the commit in $commit"
+    number=""
 else
     echo "HEAD is a simple commit."
+    number="-n 1"
 fi
 
 git show --quiet "$commit"
 
 echo ""
 
-testpaths=$(git log --format=%B -n 1 $commit | { grep -i "$ANCHOR" || true ;} | cut -b$(echo "$ANCHOR" | wc -c)-)
+testpaths=$(git log --format=%B $number $commit | { grep -i "$ANCHOR" || true ;} | cut -b$(echo "$ANCHOR" | wc -c)-)
 
 if [[ -z "$testpaths" ]]; then
     echo "Nothing to test in $commit."
--- a/testing/prow/gpu-operator.sh
+++ b/testing/prow/gpu-operator.sh
@@ -140,7 +140,7 @@ collect_must_gather() {
                 eval $ENV
 
                 echo "Copying must-gather results to $ARTIFACT_EXTRA_LOGS_DIR ..."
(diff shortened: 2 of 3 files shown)


kpouget:
@omertuc good for review when you're available:

* no more error in the finalizer:
```
Running exit finalizers ...
Running finalizer 'collect_must_gather' ...
Running the GPU Operator must-gather image ...
Operator image: nvcr.io/nvidia/gpu-operator@sha256:57e7be259af342ccdf281d23d79e15d474e2e6f506f84c0cdae3db2a199b3395
Copying must-gather results to /logs/artifacts/012__gpu-operator__must-gather ...
Running gpu_operator capture_deployment_state ...
Running gpu_operator capture_deployment_state ... done.
Operator versions collected.
Running finalizer './run_toolbox.py entitlement undeploy &> /dev/null' ...
```
* `test-path` can be specified in any commit of the PR instead of only the top-most one. See a sample result [there](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/299/pull-ci-openshift-psap-ci-artifacts-master-test-commit/1468662461887418368/artifacts/test-commit/test-commit/artifacts/) (no need to retrigger the test as it worked with the previous commit)

/approve

omertuc:
I've lost track of how this works, can you point me to the latest documentation of what exactly is being implemented / fixed here (what's the use case, and how it's expected to be used exactly)?

testing/test-commit.sh
```diff
@@ -26,17 +26,19 @@ commit="HEAD"
 parent=$(git log --pretty=%P -n 1 $commit)
 
 if grep -q " " <<< "$parent"; then
-    commit=$(cut -d" " -f2 <<< "$parent")
-    echo "HEAD is a merge commit. Taking the 2nd parent from $parent"
+    commit=$(sed 's/ /../g' <<< "$parent")
```
omertuc, line 29: I don't understand, isn't HEAD be the tip of the PR branch? In that case, why would that be a merge commit? Usually we expect PRs to be cleanly rebased so I woun't expect HEAD to be a merge commit

kpouget: no, I learned that when showing HEAD from PRs, the repo you get is a merge of `master` and `PR`, so that if your PR is based on an outdated `master`, you still test it as it would be after the PR is merged

that's also why PRs which `need-rebase` cannot be tested anymore, there's a conflict preventing the merge and thus github cannot prepare the commit

PR #307, opened 2022-01-05
[deploy-cluster] Add cluster-sno/sno target

kpouget:
* `[deploy-cluster] Refactor Makefile to make it more modular`

* `[deploy-cluster] Add cluster-sno/sno target`

* `[deploy-cluster] Make DIFF_TOOL (meld) optional`


kpouget:
@fcami if you want to have a look / test / review,
that's my way to create a daily (SNO) cluster with `make sno`

this subproject isn't used anywhere in the CI, so only informal review required

/approve
/override  ci/prow/gpu-operator-e2e

kpouget:
and commit dee9ccd1612ff2ee41111ff61b94894cbf1ebeaa is mostly a refactoring, it shouldn't change much the behavior, better look at the PR commit by commit instead of looking at the full diff, will make more sense

subprojects/deploy-cluster/config_install.mk
```diff
@@ -0,0 +1,87 @@
...
+	@[[ -f ${CLUSTER_PATH}/install-config.yaml ]] && cp ${CLUSTER_PATH}/install-config{,.back}.yaml || echo
+	${OPENSHIFT_INSTALLER} create manifests --dir=${CLUSTER_PATH} > /dev/null;
+
+install: has_installer
+	@[[ -f ${CLUSTER_PATH}/install-config.yaml ]] && cp ${CLUSTER_PATH}/install-config{,.back}.yaml || echo
+	@if [ -e ${CLUSTER_PATH}/install.log ]; then \
+	  echo "INFO: Found ${CLUSTER_PATH}/install.log" \
+	  echo "ERROR: Cluster already installed ..."; \
+	  exit 1; \
+	fi;
```
jmencak, line 25: Do we need `;` here?  Similarly elsewhere at the "top" level.

subprojects/deploy-cluster/config_install.mk
```diff
@@ -0,0 +1,87 @@
...
+	time ${OPENSHIFT_INSTALLER} create cluster --dir=${CLUSTER_PATH} --log-level=debug 2>&1 | tee ${CLUSTER_PATH}/install.log
+
+config_new_install: has_installer
+	@mkdir ${CLUSTER_PATH} -p
+	@if [ -f '${INSTALL_CONFIG}' ]; then\
```
jmencak, line 30: I'm confused about the inconsistent use of `[[` and `[` any particular reason for it?

subprojects/deploy-cluster/config_install.mk
```diff
@@ -0,0 +1,87 @@
...
+	  echo "ERROR: ${INSTALL_CONFIG} already exists ...";\
+	  exit 1;\
+	fi
+	@echo "Generating ${CLUSTER_PATH}/install-config.yaml ..."
+	${OPENSHIFT_INSTALLER} create install-config --dir=${CLUSTER_PATH} --log-level=debug
+
+config_base_install:
+	@mkdir ${CLUSTER_PATH} -p
+	@if [ ! -f ${CLUSTER_PATH}/install-config.yaml ]; then \
+	  if [ ! -f ${BASE_INSTALL_CONFIG} ]; then \
```
jmencak, line 40: It would be nice to have consistent quoting when referring to variables especially in conditions like this one.  I.e. `[ ! -f "${BASE_INSTALL_CONFIG}" ]`  similarly elsewhere.  From my experience, it is better to quote rather than not.  Similarly elsewhere.

kpouget: fixed, should be consistent in all the subproject now

> From my experience, it is better to quote rather than not. Similarly elsewhere.

same, I think to have in mind "every path may have a space" ... but sometimes I forgot because my paths never have spaces 🙃 

the trick is to properly quote globs,

```
PREFIX=...
ls "${PREFIX}*" # doesn't work
ls "${PREFIX}"* # work

ls "${PREFIX}_{a,b}* # doesn't work
ls "${PREFIX}"_{a,b} # works
```

subprojects/deploy-cluster/entitlement.mk
```diff
@@ -0,0 +1,13 @@
+manifest_entitle:
+	@echo "PEM:	 ${ENTITLEMENT_PEM}"
+	@echo "RHSM:	 ${ENTITLEMENT_RHSM}"
+	@echo "Template: ${ENTITLEMENT_TEMPLATE}"
+	@echo "Dest:	 ${ENTITLEMENT_DST_BASENAME}*"
+	@cat ${ENTITLEMENT_TEMPLATE} \
+	  | sed "s/BASE64_ENCODED_PEM_FILE/$(shell base64 -w 0 ${ENTITLEMENT_PEM})/g" \
+	  | sed "s/BASE64_ENCODED_RHSM_FILE/$(shell base64 -w 0 ${ENTITLEMENT_RHSM})/g" \
+	  > ${ENTITLEMENT_DST_BASENAME}.yaml
+	@awk '{ print > "${ENTITLEMENT_DST_BASENAME}_"++i".yaml" }' RS='---\n' ${ENTITLEMENT_DST_BASENAME}.yaml
```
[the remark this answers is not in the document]
kpouget: not sure where I got that from, must be SO ;-)

I added a comment to explain what is done, just in case
```
	# Split "${ENTITLEMENT_DST_BASENAME}.yaml" into multiple files containing a single YAML object
	# openshift-install doesn't allow files with multiple objects
```

kpouget (commented):
thanks for the feedback @jmencak, I followed the suggestions

kpouget:
I added two commits to 

1. make `manifest_entitle` target conditioned by the definition of `ENTITLEMENT_PEM`. If this path isn't defined, `make manifest_entitle` becomes a nop target.
2. add `make manifest_entitle` to `make cluster_sno`

subprojects/deploy-cluster/config_install.mk
```diff
@@ -0,0 +1,92 @@
+# https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-install-linux.tar.gz"
+# https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp-dev-preview/pre-release/openshift-install-linux.tar.gz"
+
+has_installer: ${OPENSHIFT_INSTALLER}
+${OPENSHIFT_INSTALLER}:
+	@echo "WARNING: Installer v${OCP_VERSION} not found: ${OPENSHIFT_INSTALLER}"
+	@echo "Downloading it from: "
+	mkdir -p $$(dirname  "${OPENSHIFT_INSTALLER}")
+	wget "${OPENSHIFT_INSTALLER_URL}" && set -x \
+        && tar xzf openshift-install-linux-${OCP_VERSION}.tar.gz openshift-install \
```
jmencak, line 10: You may want to replace the leading spaces with a tab here.

jmencak:
Thank you for the changes, Kevin.  Looks much cleaner to me now.  Looks good to me apart from the leading spaces in the Makefile.

kpouget:
thanks Jiri, fixed

jmencak:
Thank you for the changes, Kevin.
/lgtm

kpouget:
thanks @jmencak :)

/override ci/prow/gpu-operator-e2e

PR #309, opened 2022-01-07
[sno-snapshot] New subproject for OpenShift SNO Snapshot/Rollback with AWS

kpouget:
This WIP sub-project aims at providing a snapshot/rollback capability to a single-node OpenShift cluster running on AWS.

Snapshotting
------------

```
./snapshot.py

    Creates the snapshot of the current state of the SNO node, if no
    snapshot already exists.

./snapshot.py --wait

    Creates or awaits for the completion of the current AWS
    snapshot. This takes several minutes to complete, but 1/ the
    cluster is available during the snapshot creation and 2/ any
    changes performed after the snapshot creation time will not be
    included in the snapshot.

./snapshot.py --delete

    Deletes the AWS snapshot associated with this cluster, so that a
    new one can be created.
```

*Open questions*:

- Should the node be turned off before triggering the snapshot?

*Todo*:

- Include a human-readable timestamp in the snapshot name and printed in (instead of the snapshot ID)
- Delete all the OpenShift/Kubernetes objects backed by AWS objects (extra machines, routes, auto-scaling, ...)
  - These objects will be deleted during the cluster tear down anyway, but better avoid letting them dangling if we can.
   

Rolling Back
------------

```
./rollback.py

    Rolls back the cluster to the snapshot time.
    Fails if there is exising snapshot.
    Fails if the snapshot isn't completed yet.
    Waits for the completion of the AWS snapshot rollback and the
    reloading of RHCOS/OpenShift.

./rollback.py --wait-snapshot

   Wait for the completion of the AWS snapshot if it is not ready.

./rollback.py --only-wait-openshift

  Only wait for OpenShift to respond properly.
```

### DIFF
--- a/subprojects/sno-snapshot/README.md
+++ b/subprojects/sno-snapshot/README.md
@@ -1,0 +1,63 @@
+OpenShift SNO Snapshot/Rollback with AWS
+========================================
+
+This WIP sub-project aims at providing a snapshot/rollback capability
+to a single-node OpenShift cluster running on AWS.
+
+Snapshotting
+------------
+
+```
+./snapshot.py
+
+    Creates the snapshot of the current state of the SNO node, if no
+    snapshot already exists.
+
+./snapshot.py --wait
+
+    Creates or awaits for the completion of the current AWS
+    snapshot. This takes several minutes to complete, but 1/ the
+    cluster is available during the snapshot creation and 2/ any
+    changes performed after the snapshot creation time will not be
+    included in the snapshot.
+
+./snapshot.py --delete
+
+    Deletes the AWS snapshot associated with this cluster, so that a
+    new one can be created.
+```
+
+*Open questions*:
+
+- Should the node be turned off before triggering the snapshot?
+
+*Todo*:
+
+- Include a human-readable timestamp in the snapshot name and printed
+  in (instead of the snapshot ID)
+- Delete all the OpenShift/Kubernetes objects backed by AWS objects
+  (extra machines, routes, auto-scaling, ...)
+  - These objects will be deleted during the cluster tear down anyway,
+    but better avoid letting them dangling if we can.
+
+Rolling Back
+------------
+
+```
+./rollback.py
+
+    Rolls back the cluster to the snapshot time.
+    Fails if there is exising snapshot.
+    Fails if the snapshot isn't completed yet.
+    Waits for the completion of the AWS snapshot rollback and the
+    reloading of RHCOS/OpenShift.
+
+
+./rollback.py --wait-snapshot
+
+   Wait for the completion of the AWS snapshot if it is not ready.
+
+./rollback.py --only-wait-openshift
+
+  Only wait for OpenShift to respond properly.
+```
--- a/subprojects/sno-snapshot/src/common.py
+++ b/subprojects/sno-snapshot/src/common.py
@@ -1,0 +1,114 @@
+import time, datetime
+
+print("Importing OpenShift/Kubernetes packages ...")
+
+import kubernetes
+import ocp_resources
+import openshift
+
+from ocp_resources.node import Node
+from ocp_resources.machine import Machine
+from ocp_resources.node import Node
+
+from openshift.dynamic import DynamicClient
+try:
+    client_k8s = DynamicClient(client=kubernetes.config.new_client_from_config())
+except Exception:
+    client_k8s = None
+    print("WARNING: kubernetes not available.")
+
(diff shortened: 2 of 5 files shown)


subprojects/sno-snapshot/README.md
```diff
@@ -0,0 +1,63 @@
...
+    but better avoid letting them dangling if we can.
+
+Rolling Back
+------------
+
+```
+./rollback.py
+
+    Rolls back the cluster to the snapshot time.
+    Fails if there is exising snapshot.
```
jmencak, line 50: `s/is exising/is an existing/` ?

subprojects/sno-snapshot/src/snapshot.py
```diff
@@ -0,0 +1,83 @@
...
+    print(f"Volume Id: {volume.id}")
+
+    ts = common.human_ts()
+
+    snapshot = volume.create_snapshot(
+        Description=f'Snapshot of cluster {cluster_name}.',
+        TagSpecifications=[{
+            'ResourceType': 'snapshot',
+            'Tags': [
+                {'Key': 'Name', 'Value': f'{cluster_name}-{ts}-snapshot'},
```
jmencak, line 23: I admit I'm heavily influenced by (the beauty of) ZFS, so I'd like to see something like `@{ts}` instead of this for a snapshot, but your preference may vary.

subprojects/sno-snapshot/src/common.py
```diff
@@ -0,0 +1,114 @@
...
+            prev = snapshot.progress
+
+        time.sleep(10)
+        snapshot.reload()
+        if prev != snapshot.progress:
+            prev = snapshot.progress
+            print(snapshot.progress)
+
+def human_ts():
+    return datetime.datetime.now().strftime("%Hh%M")
```
jmencak, line 114: If possible, I'd like to see something closer to the ISO 8601 standard.  Would `%Y-%m-%dT%H:%M` work or does it include characters not allowed in shapshot names?

jmencak (commented):
Left some initial thoughts.  This is a great idea, Kevin.  Haven't tried this yet, but planning to!

subprojects/sno-snapshot/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+./snapshot.py --delete
+
+    Deletes the AWS snapshot associated with this cluster, so that a
+    new one can be created.
+```
+
+*Open questions*:
+
+- Should the node be turned off before triggering the snapshot?
```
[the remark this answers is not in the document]
kpouget: the disk snapshot should be atomic indeed, but how often Kubernetes state is persisted to the disk?

I think doing the snapshot with the node running is like removing the power plug: anything not flushed to the disk is lost, and the node reboots with this disk, potentially in a good state, potentially broken. Don't you think so?

subprojects/sno-snapshot/README.md
kpouget: sure

subprojects/sno-snapshot/src/common.py
kpouget: changed as suggested, I'll test it to be sure it doesn't fail

subprojects/sno-snapshot/src/snapshot.py
kpouget: hum, `{cluster_name}` must remain, because the names are for the whole AWS account, so better have them fully qualified
the `-snapshot` suffix can go away, as it's already the resource type

subprojects/sno-snapshot/README.md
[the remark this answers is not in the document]
jmencak: > I think doing the snapshot with the node running is like removing the power plug: anything not flushed to the disk is lost, and the node reboots with this disk, potentially in a good state, potentially broken. Don't you think so?

Yes, this is exactly what it is and how I think about it too.  Now, if you have a power failure and your system doesn't recover automatically from such a trivial problem, isn't this a fault of the system?  Would you want to use such a system?  Of course, some loss of data is expected and a snapshof of a cleanly shut down system is safer and possibly even preferred if you have time to do this and do not care about service outage while your system is shut down.

subprojects/sno-snapshot/src/snapshot.py
jmencak: I didn't want to get rid of the `{cluster_name}`, that's fine.  Just the `-snapshot` seemed a bit redundant.

kpouget:
merging

PR #317, opened 2022-01-19
Do not use deprecated app=prometheus label (#6)

kpouget:
Cherry picked from https://github.com/rh-ecosystem-edge/ci-artifacts/pull/6
```
Author: Mateusz Kowalski
```
---
This commit changes references from the `app=prometheus` label to the
`prometheus=k8s`. This is because prometheus-operator deprecates the
`app` label in 0.48.0 and removes it in 0.50.

As a consequence, the current label filter does not work anymore in OCP
4.10. We are moving to `prometheus=k8s` as this is the one that is
present in OCP 4.6 and later, which are supported releases for the GPU
Operator.

Relates-to: https://github.com/openshift/cluster-monitoring-operator/pull/1388

### DIFF
--- a/roles/gpu_operator_wait_deployment/tasks/metrics.yml
+++ b/roles/gpu_operator_wait_deployment/tasks/metrics.yml
@@ -67,7 +67,7 @@
     shell:
       set -o pipefail;
       bash "{{ gpu_operator_fetch_pod_metrics_script }}"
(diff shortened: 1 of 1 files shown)


kpouget:
merged in rh-ecosystem-edge#6,
merging here.

PR #318, opened 2022-01-19
[deploy-cluster] Deploy entitlement on master and workers on SNO

kpouget:
SNO clusters have a `master,worker` node, which requires a `role:
master` to receive MachineConfigs.

If we scale up the cluster with new nodes, they'll only have the `worker` role, so we need both `master` and `worker` objects to cover them all.


kpouget:
tested locally since several days,

merging

PR #321, opened 2022-01-27
Fix ansible-linter warnings

kpouget:
This PR fixes the few ansible-lint warning that were introduced while the linter Action was down.

They all relate to this [rule](https://ansible-lint.readthedocs.io/en/latest/default_rules.html#var-naming):

```
var-naming

All variables should be named using only lowercase and underscores
```


kpouget:
approved and lgtm in https://github.com/rh-ecosystem-edge/ci-artifacts/pull/12, 
will merge here when ^^^ is merged

PR #323, opened 2022-02-01
utils_build_push_image: improve to ease reuse

kpouget:
- prefix non-local variable names with the role name (so that there is
no clash when using multiple roles in the same playbook)

- name source artifacts with '00x_' prefix, to ease re-execution from
outside of ci-artifacts (eg, reuse the CI files to create the image
locally)

- remove 'quay.io' references, to make the code more generic

- add checkpoints to avoid rebuilding images if they already exist

- update variable names to match the BuilConfig field names

- check if the image building Pod goes out of the Pending state within
1 minute (it can stay Pending if no Node has enough memory)

- allow not providing a remote repo, to only build the image locally

- add a 'namespace' flag to specify the namespace in which the image
will be built
```
test-path: recipes run_cocodataset_nvidiadl_ssd
```
/cc @Maxusmusti


roles/utils_build_push_image/tasks/build_image.yml
```diff
@@ -0,0 +1,83 @@
...
+         {{ builconfig_name_cmd.stdout }}
+         > {{ artifact_extra_logs_dir }}/image-build.descr
+    failed_when: false
+
+  - name: Store the logs of image build (debug)
+    shell:
+      oc logs {{ build_name_cmd.stdout }}
+        -n "{{ utils_build_push_image_namespace }}"
+        > {{ artifact_extra_logs_dir }}/image-build.log
+    failed_when: false
```
Maxusmusti, line 83: Missing newline at end of file

roles/utils_build_push_image/tasks/main.yml
```diff
@@ -1,169 +1,47 @@
...
+  command:
+    oc get
+       "istag/{{ utils_build_push_image_local_name }}:{{ utils_build_push_image_tag }}"
+       -n "{{ utils_build_push_image_namespace }}"
+  register: has_image
   failed_when: false
 
-- name: Create push pod and run image push
-  command: oc apply -f {{ artifact_extra_logs_dir }}/helper-image-push-pod.yml
+- name: Build the image
```
Maxusmusti, line 27: Do we not still want to try building the helper image before attempting to build the actual image?

Maxusmusti (changes requested):
All look like great changes, just noticed a couple minor things worth noting

roles/utils_build_push_image/tasks/main.yml
kpouget: I'm not sure any more how we discussed it, but from my POV, best is:
1. build the user image, it may take time and/or fail
2. if ^^^ is successful, and we want to push, build the helper image (should never fail)
3. if ^^^ is successful, and we want to push, run the Pod to push the image

I noticed you had 1 and 2 swapped, let me know if you see it differently!

Maxusmusti: Ok yeah, given that 2 should never fail (and that pushing is optional), I think your order makes the most sense

Maxusmusti (approved):

kpouget:
@Maxusmusti can you try to `/lgtm`?

kpouget:
thanks @Maxusmusti :)

PR #330, opened 2022-02-11
[wdm] Improve

kpouget:
* `[wdm] Delete ansible temp file after execution`

---

* `[wdm] remove hardcoded path from ansible-playbook`

---

* `[wdm] flush stdout/err before calling subprocess.run`

---


kpouget:
merging

PR #335, opened 2022-02-28
Changing COCO Source to S3 Bucket and Adding Failure Conditions

Maxusmusti:
NOTE - test-path: prow benchmarking use_private_s3 download_twice benchmark_twice


Maxusmusti:
Initial test to see if successful local changes translate appropriately to CI

roles/benchmarking_deploy_coco_dataset/files/entrypoint.sh
```diff
@@ -5,107 +5,19 @@ set -u
 set -o pipefail
 set -x
 
-cd /storage
```
kpouget, line 8: @Maxusmusti, please keep this part from the original script, 
we want the download to work when we don't provide our private S3 secret

roles/benchmarking_deploy_coco_dataset/files/entrypoint.sh
```diff
@@ -5,107 +5,19 @@ set -u
...
-        continue
-    fi
-    echo "Extracting $dataset_filename ..."
-    unzip "$dataset_filename" >/dev/null
-
-    touch "${dataset_filename}.extracted"
-    echo "Deleting $dataset_filename ..."
-    rm -f "$dataset_filename"
-done
+/usr/local/bin/aws s3 cp s3://psapcocobucket/ /storage --profile=cocobucket --recursive
```
kpouget, line 16: can you replace `/storage` by an env variable, that you defined in the Pod template? (eg, `STORAGE_DIR`)
(will make it easier to reuse the script, as when Anisha wanted to download the dataset on RHEL)

roles/benchmarking_deploy_coco_dataset/files/entrypoint.sh
```diff
@@ -5,107 +5,19 @@ set -u
...
 
 echo "All done!"
```
kpouget, line 18: even with S3, we _need_ a way to validate the content of the storage directory, and not download twice if the content is valid.

maybe counting the number of file in each of the directories would be enough, assuming that S3 download cannot corrupt files. Think that `aws s3 cp` may be interrupted (Pod killed) in the middle of the download.

also, can you download the different files (or extracted directories) one by one? so that if we add more content in the bucket (eg, a preprocessed version of the files), we do not download it as well

roles/benchmarking_deploy_coco_dataset/files/entrypoint.sh
```diff
@@ -5,107 +5,19 @@ set -u
...
-for dataset_filename in $(cat checksum.sha256sum | cut -d' ' -f3);
-do
-    if [[ -f "$dataset_filename".extracted ]]; then
-        echo "$dataset_filename already downloaded and extracted, skipping."
-        continue
-    fi
+curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+unzip awscliv2.zip
+./aws/install
+/usr/local/bin/aws configure import --csv file://${CERT_FILE}
```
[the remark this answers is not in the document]
kpouget: and, please, use a bucket in PSAP account, we really do not want the bucket to be in another account, for cost management, and resource tracking

roles/benchmarking_deploy_coco_dataset/templates/pod.yml.j2
```diff
@@ -19,12 +19,12 @@ spec:
 {% endif %}
 {% if benchmarking_coco_dataset_client_cert|length %}
     - name: CERT_FILE
```
kpouget, line 21: can you rename this env into `CREDS_FILE`?

roles/benchmarking_run_nvidiadl_ssd/files/entrypoint.sh
```diff
@@ -27,7 +28,7 @@ if [ ! -f ${DATASET_DIR}/annotations/bbox_only_instances_val2017.json ]; then
         "${DATASET_DIR}/annotations/instances_val2017.json" \
         "${DATASET_DIR}/annotations/bbox_only_instances_val2017.json"
 fi
-
+comment
```
kpouget, line 31: junk

roles/benchmarking_run_nvidiadl_ssd/tasks/main.yml
```diff
@@ -77,3 +77,21 @@
...
     failed_when: false
+
+- name: Get average sample rate
+  shell:
+    tail -n9 "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.log" | head -n1 | awk '{ print $NF }'
+  register: samplerate
+
+- name: Fail if sample rate is below expectations
+  when: samplerate.stdout | float < 270.0
+  fail: msg="The average samples processed per second was unexpectedly low"
```
kpouget, line 88: (I know it's WIP), can you make this `270.0` a role parameter?
and I'm not actually sure where we want to do this speed check, 
here, or in the calling script

usually, I like to have the role general purpose, and let the calling script do the "error" checking:
270.0 value is directly linked with the INSTANCE_TYPE you chose, and maybe the threshold/epoch values

Maxusmusti:
Set up a new dataset bucket in psap org (via my guide) and switched everything to use it

Maxusmusti:
Updated and cleaned up the download role, can now do both s3 and original download, and can also specify storage dir

Maxusmusti:
Tested both s3 and original download (as well as cache checks)

Maxusmusti:
Made corrections to cache checks and made sure benchmark equally functional with both download methods

Maxusmusti:
I've abstracted and generalized the benchmark evaluation process. Question @kpouget, what would you like to change the name `nvidiadl` to?

roles/benchmarking_run_nvidiadl_ssd/tasks/main.yml
```diff
@@ -77,3 +77,15 @@
       oc describe pod/{{ benchmarking_nvidiadl_ssd_name }} -n {{ benchmarking_namespace }}
         > "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.descr"
     failed_when: false
+
+- name: Get average sample rate
+  shell:
+    set -o pipefail;
+    cat "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.log" | grep avg. | tail -n1 | awk '{ print $NF }' > "{{ artifact_dir }}/sample_rate.log"
```
kpouget, line 84: 🤔 
I wasn't thinking about it this way, but it might be a good idea
I was thinking: write it into `{{ artifact_extra_logs_dir }}`, and fetch it from there in the calling script, but it makes sense like that as well.

Could you 
1/ prefix the file name with the role: `"{{ artifact_dir }}/benchmarking_run_ssd_sample_rate.log"`
2/ write the file as well in `{{ artifact_extra_logs_dir }}`

for both files of course

roles/benchmarking_run_nvidiadl_ssd/tasks/main.yml
```diff
@@ -77,3 +77,15 @@
...
+  register: samplerate
+
+- name: Get time to run benchmark
+  shell:
+    set -o pipefail;
+    cat "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.log" | grep real | awk '{ print $NF }' | awk -Fm '{print $1}' > "{{ artifact_dir }}/bench_duration.log"
+  register: benchduration
```
kpouget, line 91: register: samplerate` and `register: benchduration` not used, can be removed

testing/recipes/run_cocodataset_nvidiadl_ssd.sh
```diff
@@ -51,4 +53,5 @@ fi
 
 for i in $(seq $RUN_CNT); do
     ./run_toolbox.py benchmarking run_nvidiadl_ssd "$gpu_node_hostname" --epochs=$EPOCHS --threshold=$THRESHOLD
+    ./run_toolbox.py benchmarking assess_benchmark_stats --min_sample_rate=$MINSR --max_duration=$MAXDR
```
kpouget, line 56: I don't think this requires a new role, just write a bash function  doing the same checks as what you do in `assess_benchmark_stats`

and actually, it may be time to add a new script `testing/prow/benchmarking.sh` that will do something similar to this one + the performance assessement

in this script, just let everything else as you've modified here, for easy reexecution

playbooks/benchmarking_assess_stats.yml
```diff
@@ -0,0 +1,7 @@
+---
```
kpouget, line 1: mentioned in another comment, no need for a dedicated role

roles/benchmarking_deploy_coco_dataset/files/entrypoint.sh
```diff
@@ -5,107 +5,123 @@ set -u
...
-    echo "INFO: no cert file provided, downloading from upstream."
+    ann_check=$(ls -1q ${STORAGE_DIR}/annotations | wc -l)
+    train_check=$(ls -1q ${STORAGE_DIR}/train2017 | wc -l)
+    val_check=$(ls -1q ${STORAGE_DIR}/val2017 | wc -l)
+
+    if [[ $ann_check != 8 ]]; then
+        rm -rf ${STORAGE_DIR}/annotations
+        aws s3 cp s3://psap-coco-bucket/annotations/ ${STORAGE_DIR}/annotations --profile=cocobucket --recursive
+    fi
+    if [[ $train_check != "118287" ]]; then
```
kpouget, line 38: can you write the expected number of files in a bash array, to make it clearer what the values mean?
eg,
```
EXPECTED_NUMBER_OF_FILES=(associative array)
EXPECTED_NUMBER_OF_FILES[annotations] = 8
EXPECTED_NUMBER_OF_FILES[train2017] = 118287
...
```
you can even loop across all the keys and 
```

        rm -rf ${STORAGE_DIR}/$key
        aws s3 cp s3://psap-coco-bucket/$key/ ${STORAGE_DIR}/$key --profile=cocobucket --recursive
    fi
```
and variables must be protected by `"` as often as possible,
imagine there's a space in `STORAGE_DIR="/ stupid dirname -rf help/"`
this will execute `rm -rf / stupid dirname -rf help/`
instead of simply failing because the directory doesn't exist

kpouget (changes requested):
would you be able to split the coco reword | failure condition in two differents commits or even PR?
https://github.com/mystor/git-revise is very powerful for that,
but here `git` should be enough,
```
git reset master # remove all the commits (but not their content)
git add -p # add files related to the coco download
git commit -m "coco from s3"
git add -p # add files related to the failure
git commit -m "failure"
```

Maxusmusti:
Applied all code review feedback and split into two distinct commits

testing/prow/benchmarking.sh
```diff
@@ -0,0 +1,78 @@
...
+        echo "Sample rate test passed!"
+    fi
+    if [[ $MAXDR -lt $bench_duration ]] ; then
+        echo "Benchmark duration above maximum expected"
+        return 0
+    else
+        echo "Benchmark duration test passed!"
+    fi
+    return 1
+}
```
kpouget, line 30: by convention, `0` means "everything's OK", and `not 0` means that an error occur

so here it would make more sense to have `return 1` in case of error.

Can you quote `$bench_duration` and `$sample_rate`? (mandatory as read from another file)
and `$MAXDR` + `$MINSR` for consistancy (not "mandatory", as we can visually asses its value)

testing/prow/benchmarking.sh
```diff
@@ -0,0 +1,78 @@
...
+assess_benchmark_stats () {
+    artifact_dir=${ARTIFACT_DIR:-"/tmp/ci-artifacts_$(date +%Y%m%d)"}
+    sample_rate=$(cat "${artifact_dir}/benchmarking_run_ssd_sample_rate.log")
+    bench_duration=$(cat "${artifact_dir}/benchmarking_run_ssd_bench_duration.log")
+    if [[ $MINSR -gt $sample_rate ]] ; then
+        echo "Sample rate below minimum expected"
+        return 0
+    else
+        echo "Sample rate test passed!"
+    fi
```
kpouget, line 22: you can remove the `else`, that's a "guard" pattern:

```
if not test:
  return err
...
```

which avoid multiple levels of indentation (in the worst case):
```
if test:
   ... 
   do 
   many 
   things

return 0
```
or
```
if test:
  do something
    if test2:
       do something2

return
```
this one can be rewritten with all the important code at level 0

testing/prow/benchmarking.sh
```diff
@@ -0,0 +1,78 @@
...
+THRESHOLD=0.05
+MINSR=150.0
+MAXDR=30
+
+assess_benchmark_stats () {
+    artifact_dir=${ARTIFACT_DIR:-"/tmp/ci-artifacts_$(date +%Y%m%d)"}
+    sample_rate=$(cat "${artifact_dir}/benchmarking_run_ssd_sample_rate.log")
+    bench_duration=$(cat "${artifact_dir}/benchmarking_run_ssd_bench_duration.log")
+    if [[ $MINSR -gt $sample_rate ]] ; then
+        echo "Sample rate below minimum expected"
```
kpouget, line 18: in the `echo` of this function (or maybe before the tests), can you print the content of `$MINSR ` and `$sample_rate` (and same for the duration)

otherwise, when trouble shooting, you look at the logs, see:
```
ERROR: Sample rate below minimum expected
```
(can you add the `ERROR` prefix) and say: great, it was too low, but how low was it? and what was the limit?

Bash's `set -x` would tell, but better make it obvious and state in plain text what went wrong

testing/prow/benchmarking.sh
```diff
@@ -0,0 +1,78 @@
...
+        return 0
+    else
+        echo "Benchmark duration test passed!"
+    fi
+    return 1
+}
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+PSAP_SECRET_PATH=/var/run/psap-entitlement-secret
```
kpouget, line 34: can you move this above `EPOCHS=3`? (constants always/often go first)

kpouget (changes requested):
thanks @Maxusmusti , looks good :)

I added multiple comments, but they are really minor, just moving code bits up and down

roles/benchmarking_run_nvidiadl_ssd/tasks/main.yml
```diff
@@ -77,3 +77,15 @@
       oc describe pod/{{ benchmarking_nvidiadl_ssd_name }} -n {{ benchmarking_namespace }}
         > "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.descr"
     failed_when: false
+
+- name: Get average sample rate
+  shell:
+    set -o pipefail;
+    cat "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.log" | grep avg. | tail -n1 | awk '{ print $NF }' > "{{ artifact_dir }}/benchmarking_run_ssd_sample_rate.log";
```
kpouget, line 84: can you add a unit here, so that the file content is easier to read? (mind that the asses function should be updated as well)
```
awk '{ print $NF " units" }'
```

roles/benchmarking_run_nvidiadl_ssd/tasks/main.yml
```diff
@@ -77,3 +77,15 @@
...
+    cp {{ artifact_dir }}/benchmarking_run_ssd_sample_rate.log {{ artifact_extra_logs_dir }}/benchmarking_run_ssd_sample_rate.log
+
+- name: Get time to run benchmark
+  shell:
+    set -o pipefail;
+    cat "{{ artifact_extra_logs_dir }}/pod_run-nvidiadl-ssd.log" | grep real | awk '{ print $NF }' | awk -Fm '{print $1}' > "{{ artifact_dir }}/benchmarking_run_ssd_bench_duration.log";
```
kpouget, line 90: same here for adding the unit

testing/prow/benchmarking.sh
```diff
@@ -0,0 +1,75 @@
...
+EPOCHS=3
+THRESHOLD=0.05
+MINSR=150.0
+MAXDR=30
+
+assess_benchmark_stats () {
+    artifact_dir=${ARTIFACT_DIR:-"/tmp/ci-artifacts_$(date +%Y%m%d)"}
+    sample_rate=$(cat "${artifact_dir}/benchmarking_run_ssd_sample_rate.log")
+    bench_duration=$(cat "${artifact_dir}/benchmarking_run_ssd_bench_duration.log")
```
kpouget, line 17: mind here for the unit, I guess you can add `| cut -d" " -f1` to remove the unit

testing/prow/benchmarking.sh
```diff
@@ -0,0 +1,75 @@
...
+    exit 1
+else
+    echo "Using GPU node name: $gpu_node_hostname"
+fi
+
+DL_OPT=""
+if [[ "$@" == *use_private_s3* ]]; then
+    S3_CRED="${PSAP_SECRET_PATH}/cococred.csv"
+    DL_OPT="--s3_cred=$S3_CRED"
+fi
```
kpouget, line 52: can you get the test running with `use_private_s3 download_twice benchmark_twice` to validate all of that?

kpouget (changes requested):
thanks @Maxusmust, almost there,
only need to add the units in the speed files,
and test it with more options to validate S3 download, download twice, and benchmark twice

testing/prow/benchmarking.sh
Maxusmusti: How exactly do I add those arguments for the CI test? Should I add functionality for a comment similar to `test-path: prow benchmarking`, or should the args be specified directly in some file like run or test-commit (or just directly in benchmarking.sh)?

Maxusmusti:
Added units

kpouget:
```
++ ls -1q /storage/train2017
++ wc -l
ls: cannot access '/storage/train2017': No such file or directory
+ check=0
```
better:
```
(ls -1q /storage/train2017 2>/dev/null || true) | wc -l
```
and while you modify it, can you mute `unzip awscliv2.zip`? either with `unzip awscliv2.zip >/dev/null` (not sure if it's stdout or stderr), or maybe there is an `unzip` flag to be less verbose (would be better)

kpouget:
thanks @Maxusmusti, looks great :)

the only think that still have to be changed is `s/nvidiadl/mlperf`:
* role name
* file names (`pod_run-nvidiadl-ssd.*`)
* calling scripts

at the same time, can you rename `testing/prow/benchmarking.sh` into `testing/benchmarking/run_mlperf_ssd.sh`

and get it tested with
``
test-path: benchmarking run_mlperf_ssd use_private_s3 download_twice benchmark_twice
```

and maybe we can have a quick call together, so that I can show you the easiest ways to merge the 4+ commits into two?

**Comment by kpouget** [comment]
thanks @Maxusmusti, 👍🏻  to merge \o/
we'll discuss together later today to properly squash the commits into 2 commits to have a clean git history

**Comment by kpouget** [comment]
@Maxusmusti you need to test with:

```
test-path: benchmarking run_mlperf_ssd ...(args)...
```

the error message is a bit off leading, 
```
FATAL: Unknown entrypoint 'prow' | benchmarking use_private_s3 download_twice benchmark_twice
```
it should read like this, if you want to change it:
```
FATAL: Unknown entrypoint 'benchmarking' inside 'prow' directory | benchmarking use_private_s3 download_twice benchmark_twice
```

**Comment by Maxusmusti** [comment]
yeah, just noticed I had used the old test-path, updated error message as well, retesting now

**Comment by kpouget** [comment]
```
 The connection to the server api.ci-ocp-4-10-amd64-aws-us-east-1-tlbhj.hive.aws.ci.openshift.org:6443 was refused - did you specify the right host or port? 
```
cluster was destroyed mid-flight :'(

/test test-commit

**Comment by kpouget** [comment]
hum :/
```
  Warning  FailedCreate  21m   awscontroller  ci-ocp-4-10-amd64-aws-ndnh4-worker-us-west-1a--p3-2xlarge-v2zcs: reconciler failed to Create machine: failed to launch instance: error launching instance: The requested configuration is currently not supported. Please check the documentation for supported configurations.
```
but the instance type exists, and is available in this [region](https://aws.amazon.com/ec2/instance-types/p3/).
Maybe 4.10 doesn's support it?

/test 4.9-test-commit

**Comment by Maxusmusti** [comment]
`TASK: cluster_set_scale : Wait for all machinesets with type p3.2xlarge to be ready`
`----- FAILED ----`

@kpouget do we have to make an adjustment elsewhere to have a p3.2xlarge node ready?

**Comment by kpouget** [comment]
```

oc get packagemanifests/nfd -n openshift-marketplace > /dev/null
----
error: the server doesn't have a resource type "packagemanifests"
----
```
🤦🏻 
flake
/test 4.9-test-commit

Maxusmusti:
@kpouget seems like the same issue from 4.10 occurred in 4.9 w.r.t. the p3.2xlarge instance

kpouget:
thanks @Maxusmusti !

merging!

PR #338, opened 2022-03-07
cluster_deploy_operator: fix typo in the subscription name

kpouget:
Note that this typo did not affect the good behavior of the role, as
long as no two operators were deployed in the same namespace.


kpouget:
live in https://github.com/rh-ecosystem-edge/ci-artifacts/,
merging here

PR #358, opened 2022-04-26
Remove unused/outdated roles

kpouget:
/cc @kpouget


kpouget:
@dagrayvid no objection on removing SRO related roles, correct?

dagrayvid:
> @dagrayvid no objection on removing SRO related roles, correct?

Looks good to me.

PR #403, opened 2022-06-30
rhods_test_jupyterlab: test-jupyterlab-psap-simplenotebook: remove 'Clean up the files and folders we created'

kpouget:
This PR removes the `Clean up the files and folders we created` step from the simple test robot.
This steps often fails, and it isn't relevant for us (we delete the PVCs after the testing).


fcami (approved):
Makes total sense to me.

kpouget:
> Makes total sense to me.

good, merging :)

PR #415, opened 2022-07-06
testing: ods: common: properly disable ENABLE_AUTOSCALER

kpouget:
This PR properly disables the auto-scaler enablement.

The previous value (`0`) enabled the auto-scaling, so I added a
comment to make it clear that not-empty is enabled.

/cc @fcami 
/approve

### DIFF
--- a/testing/ods/common.sh
+++ b/testing/ods/common.sh
@@ -48,7 +48,8 @@ OCP_WORKER_MACHINE_TYPE=m5.xlarge
 
 OCP_BASE_DOMAIN=psap.aws.rhperfscale.org
 
-ENABLE_AUTOSCALER=0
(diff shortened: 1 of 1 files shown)


testing/ods/common.sh
```diff
@@ -48,7 +48,8 @@ OCP_WORKER_MACHINE_TYPE=m5.xlarge
 
 OCP_BASE_DOMAIN=psap.aws.rhperfscale.org
 
-ENABLE_AUTOSCALER=0
+# if not empty, enables auto-scaling in the sutest cluster
+ENABLE_AUTOSCALER=
```
fcami, line 52: Approved, but I would like it better if the logic was 0 vs 1 for instance

fcami (approved):

PR #420, opened 2022-07-07
WIP: RHODS: scale test with 100 users

kpouget:
WIP PR for running scale tests with the nodes pre-provisionned.

PR #404, opened 2022-06-30
WIP: RHODS: test authentication with 300 simultaneous users

kpouget:
This WIP PR disables the `Notebook` part of the test suite (and forces the sutest node count to 1 as no notebook will be created), and launches N simultaneous users.

PR #421, opened 2022-07-07
RHODS: various updates

kpouget:
* `rhods_test_jupyterlab: entrypoint: remove call to 'oc login'`

The update to ODS-CI makes it unnecessary.

---

* `rhods_test_jupyterlab: new flag for artifacts collection: no-image-except-failed-and-zero`

This new flag allows collecting the artifacts if ...
- the test failed (unchanged, for troubleshooting)
- the test-pod index is 0 (to demonstrate what happened during the test)

---

* `testing: ods: jh-at-scale: use 'no-image-except-failed-and-zero' to collect test_jupyterlab artifacts`

Use the new flag `no-image-except-failed-and-zero` so that the
screenshots of the process 0 are collected, to demonstrate what
happens during a successful (hopefully) run.

---

* `rhods_test_jupyterlab: test-jupyterlab-psap-simplenotebook: remove useless test 'Login to Jupyterhub'`

These steps are already performed in `Begin and Authenticate`.

---

* `[deploy-cluster] Add ability to export the metadata.json to a safe place`

This commit adds an optional variable flag to export the cluster's
`metadata.json` before the cluster creation.

This is required in the CI to save the file to the shared directory
_before_ creating any resource. So if the cluster creation pod is
interrupted (and the `trap` isn't executed), the `metadata.json` is
still shared with the cluster destruction pod, and the cluster can be
properly destroyed.

---

* `testing: ods: ocp_cluster: save metadata.json as part of the cluster creation`

If the install pod is terminated, this file won't be saved, and the
cluster would be left dangling.

---

* `testing: gather-extra.sh: do not use /tmp to save the container list`

When using `/tmp/containers` from two `gather-extra.sh` running
concurrently, one file would overwrite the other, making the gathering
inconsistent.

In addition, remove the collection of `/tmp/node` and `/tmp/pods-api`
which were not used.

---

When this PR gets merged, I'll update `[jh-at-scale](https://github.com/openshift-psap/ods-ci/tree/jh-at-scale)` with the content of `[jh-at-scale-next](https://github.com/openshift-psap/ods-ci/tree/jh-at-scale-next)` (see this PR https://github.com/openshift-psap/ods-ci/pull/6, even if it cannot be merged, because a `push --force` is what we need)

--

/cc @fcami


kpouget:
The commits have been tested as part of https://github.com/openshift-psap/ci-artifacts/pull/404#issuecomment-1178025181 and https://github.com/openshift-psap/ci-artifacts/pull/420#issuecomment-1178078538.

/approve

PR #428, opened 2022-07-12
RHODS: improve get-cluster's job

kpouget:
/cc @kpouget


kpouget:
not 100% functional yet (the kubeadmin forced password doesn't work), but I'll solve it later.
Merging.

PR #433, opened 2022-07-13
RHODS: cluster_undeploy_ldap: also undeploy on osd/rosa 

kpouget:
* `cluster_undeploy_ldap: also undeploy on osd/rosa`

The role `cluster_undeploy_ldap` was only undeploying the LDAP setup
in OCP, and not in OSD/ROSA.

This commit fixes this missing point.

---

* `testing: ods: jh-at-scale: undeploy LDAP after testing`

Until now, the LDAP authentication was not undeployed in the CI
testing.

This commit fixes that, so that the SUTest cluster is left in a
cleaner state.

---


kpouget:
failed because osd cannot have a machinepool with 1 node. Fixed that, retrying.

/test ods-jh-on-osd

kpouget:
Merging, we'll confirm that it works correctly with OSD later.

PR #442, opened 2022-07-15
RHODS: wait for RHODS images to be loaded and look up in the cluster the image tag to preload

kpouget:
* `testing: ods: jh-at-scale: look up in the cluster the image tag to preload`

This part was failing on OSD, as the addon is installing the "latest"
version of RHODS and not a specific one as we do in OCP.

---

* `rhods_wait_ods: wait for RHODS images to be loaded`

---

/cc @kpouget


kpouget:
tested in https://github.com/openshift-psap/ci-artifacts/pull/420, merging.

PR #454, opened 2022-07-21
cluster_destroy_ocp: allow passing no option to delete the local cluster

kpouget:
This commit updates the `cluster_destroy_ocp` to look up the cluster's
REGION and TAG_ID with `oc` (from the `machine` AWS specs).

These two values are then passed to `openshift-install` to destroy the
cluster.

This version of the toolbox can only run if the `--confirm` flag is
passed.

---

Tested locally.

/cc @fcami


fcami (approved):
/lgtm

PR #459, opened 2022-07-25
testing: run: do not crash when the pull request data cannot be fetched

kpouget:
This breaks the rehearsals from openshift/release.

### DIFF
--- a/testing/run
+++ b/testing/run
@@ -79,8 +79,11 @@ prechecks() {
     (git describe HEAD --long --always || echo "git missing") > ${ARTIFACT_DIR}/ci_artifact.git_version
 
     if [[ "${PULL_NUMBER:-}" ]]; then
-        curl -sSf "https://api.github.com/repos/openshift-psap/ci-artifacts/pulls/$PULL_NUMBER" -o "${ARTIFACT_DIR}/pull_request.json"
-        curl -sSf "https://api.github.com/repos/openshift-psap/ci-artifacts/issues/420/comments" -o "${ARTIFACT_DIR}/pull_request-comments.json"
+        PR_URL="https://api.github.com/repos/openshift-psap/ci-artifacts/pulls/$PULL_NUMBER"
+        PR_COMMENTS_URL="https://api.github.com/repos/openshift-psap/ci-artifacts/issues/$PULL_NUMBER/comments"
+
+        curl -sSf "$PR_URL" -o "${ARTIFACT_DIR}/pull_request.json" || echo "WARNING: Failed to download the PR from $PR_URL"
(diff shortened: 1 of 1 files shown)


testing/run
```diff
@@ -79,8 +79,8 @@ prechecks() {
     (git describe HEAD --long --always || echo "git missing") > ${ARTIFACT_DIR}/ci_artifact.git_version
 
     if [[ "${PULL_NUMBER:-}" ]]; then
-        curl -sSf "https://api.github.com/repos/openshift-psap/ci-artifacts/pulls/$PULL_NUMBER" -o "${ARTIFACT_DIR}/pull_request.json"
-        curl -sSf "https://api.github.com/repos/openshift-psap/ci-artifacts/issues/420/comments" -o "${ARTIFACT_DIR}/pull_request-comments.json"
+        curl -sSf "https://api.github.com/repos/openshift-psap/ci-artifacts/pulls/$PULL_NUMBER" -o "${ARTIFACT_DIR}/pull_request.json" || true
```
jmencak, line 82: Isn't this introducing silent failures?  Shouldn't at least the failure be logged somewhere?

jmencak: Actually, I take this back (`-S`).  So fine with me. :)

kpouget: good point, but the failure isn't "silent", `curl -sSf` will log an error message in `stderr`, as at the bottom of [this log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_release/30732/rehearse-30732-periodic-ci-openshift-psap-ci-artifacts-master-ods-cleanup-velero-buckets/1550478314798321664/build-log.txt)
```
curl: (22) The requested URL returned error: 404 
```
true, that's not the clearest message (I thought `set -x` was active in this file, but it's not)

kpouget: @fcami agrees with you offline,

I'll change this to:
```
curl ... || echo "Failed to download ..."
```

testing/run
```diff
@@ -79,8 +79,8 @@ prechecks() {
...
+        curl -sSf "https://api.github.com/repos/openshift-psap/ci-artifacts/issues/420/comments" -o "${ARTIFACT_DIR}/pull_request-comments.json" || true
```
kpouget, line 83: lol, this will always work 🙃

jmencak:
> /lgtm

With a nit that I've been taught to report errors and warnings on `stderr`.

kpouget:
> With a nit that I've been taught to report errors and warnings on stderr.

one day we'll have a cleaner way than Bash to write the entrypoints of the test, and that will include proper logging :)

thanks for having a look @jmencak :)

/approve

PR #461, opened 2022-07-26
[cloud-watch] Refuse to run with --delete but without --ci-delete-older-than

kpouget:
Otherwise the command may delete all the CI clusters ...

/cc @kpouget


kpouget:
merging, still not finalized

PR #488, opened 2022-08-29
testing: ods: run_private_notebook_ux_e2e_scale_test: new entrypoint to test a private cluster

kpouget:
This entrypoint requires two extra variables to be defined when
launching the test:

```
SUTEST_CLUSTER_NAME must be set with the base name of the private cluster
SUTEST_CLUSTER_USER_NAME must be set with the username to use to log into the private cluster
```

Example:

```
/test ods-nb-ux-on-private
/var SUTEST_CLUSTER_NAME my.cluster.base.name
/var SUTEST_CLUSTER_USER_NAME my_username
```

### DIFF
--- a/testing/ods/private_cluster.sh
+++ b/testing/ods/private_cluster.sh
@@ -1,0 +1,128 @@
+#! /bin/bash
+
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o errtrace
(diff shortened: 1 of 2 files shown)


kpouget:
merging, I'll test this further in other PRs.

PR #493, opened 2022-09-05
testing: pr_args: use /var instead of /env

kpouget:
As the variable is not exposed in the environment variables.

/cc @kpouget


kpouget:
merging, I'll test this further in other PRs.

PR #496, opened 2022-09-07
RHODS: Integrate Locust-based API-level notebook load testing

kpouget:
/var NOTEBOOK_TEST_FLAVOR=api-level
/var ODS_CATALOG_IMAGE=quay.io/llasmith/rhods-operator-live-catalog
/var ODS_CATALOG_IMAGE_TAG=1.18.0-6

PR #510, opened 2022-09-19
RHODS: add API-level testing role

kpouget:
/cc @kpouget


kpouget:
tested in https://github.com/openshift-psap/ci-artifacts/pull/496, merging.

PR #519, opened 2022-10-03
cluster_deploy_operator: fix issues when deploying in all the namespaces

kpouget:
Tested locally.

/cc @kpouget


kpouget:
I will test it further with other PRs (to avoid wasting resources)

PR #564, opened 2022-11-04
Switch to OpenVINO

fcami:
/var matbench.config_file=model_mesh.yaml

### DIFF
--- a/testing/model-mesh/4-deploy-many-ns-many-models.sh
+++ b/testing/model-mesh/4-deploy-many-ns-many-models.sh
@@ -53,11 +53,12 @@ do
     if [[ "$API_ENDPOINT_CHECK" -eq 0 ]]
     then
 	NS=${NS_BASENAME}-${i}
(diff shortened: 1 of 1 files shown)


testing/model-mesh/openvino-serving-runtime.yaml
```diff
@@ -0,0 +1,63 @@
...
+  protocolVersions:
+    - grpc-v1
+  multiModel: true
+
+  grpcEndpoint: "port:8085"
+  grpcDataEndpoint: "port:8001"
+
+  containers:
+    - name: ovms
+      image: openvino/model_server:2022.2
```
kpouget, line 41:
```suggestion
      image: docker.io/openvino/model_server:2022.2
```
better with a fullly qualified image address

fcami: 100% agreed, but I won't make the change here.

PR #572, opened 2022-11-14
Model Mesh: switch to the ODH Operator

fcami:
/var matbench.config_file=model_mesh.yaml


kpouget:
thanks, looks good
/lgtm

PR #738, opened 2023-06-01
Icelake Cluster 200 user test

kdvalin:
Added tests to use for 200 users targeting the icelake cluster.  Will be updating this with information from Jenkins as it becomes available.

### DIFF
--- a/testing/notebooks/config.yaml
+++ b/testing/notebooks/config.yaml
@@ -232,6 +232,7 @@ ci_presets:
     extends: [metal]
     clusters.driver.compute.dedicated: false
     clusters.sutest.compute.dedicated: false
+    'ci_presets.notebook_performance["clusters.sutest.compute.dedicated"]': false
     'ci_presets.notebook_scale_test["tests.notebooks.users.count"]': 150
 
   spot:
@@ -249,15 +250,11 @@ ci_presets:
     tests.notebooks.users.count: 15
   
   icelake_notebook_scale_test_jenkins:
-    extends: [notebook_scale_test, metal]
-    clusters.driver.compute.dedicated: false
-    clusters.sutest.compute.dedicated: false
+    extends: [notebook_scale_test, metal, icelake]
     tests.notebooks.users.count: 200
   
   icelake_notebook_gating_test:
-    extends: [notebook_gating_scale_test, metal]
-    clusters.driver.compute.dedicated: false
-    clusters.sutest.compute.dedicated: false
+    extends: [notebook_gating_scale_test, metal, icelake]
     tests.notebooks.users.count: 200
 
 secrets:
--- a/testing/notebooks/config.yaml
+++ b/testing/notebooks/config.yaml
@@ -247,6 +247,14 @@ ci_presets:
     clusters.driver.compute.dedicated: false
     clusters.sutest.compute.dedicated: false
     tests.notebooks.users.count: 15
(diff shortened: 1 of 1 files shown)


kdvalin:
First test after fixing config/param issues:
https://ci.app-svc-perf.corp.redhat.com/job/ExternalTeams/job/RHODS/job/ci-artifacts/119/

Currently only running 5 users, updating config to use ` 'ci_presets.notebook_scale_test["tests.notebooks.users.count"]': 200`

kdvalin:
Same problem: https://ci.app-svc-perf.corp.redhat.com/job/ExternalTeams/job/RHODS/job/ci-artifacts/120/

kdvalin:
Got a 200 user test running, only 5 users failed
https://ci.app-svc-perf.corp.redhat.com/job/ExternalTeams/job/RHODS/job/ci-artifacts/123/artifact/run/f23-h33-000-6018r.rdu2.scalelab.redhat.com/002_run_tests_and_plots/001__test_run/002__plots/

Runs are being uploaded to Horreum, but no Prometheus metrics are being posted :/  Will look into

kdvalin:
Gating test did not run, it does not look like the step actually triggered.

https://ci.app-svc-perf.corp.redhat.com/job/ExternalTeams/job/RHODS/job/ci-artifacts/124/artifact/run/f23-h33-000-6018r.rdu2.scalelab.redhat.com/

kdvalin:
As for the prom metrics, looks like `only_initialize=True` for `prom.register` is not doing what it's supposed to do.. Yay

kpouget:
> icelake_notebook_gating_test:
> extends: [gating_test, metal]
strange, this typo should have triggered a failure, you shouldn't be able to apply a preset [which doesn't exist](https://github.com/openshift-psap/ci-artifacts/blob/main/testing/utils/configure.sh#L41):
```
    if [[ "$values" == null ]]; then
        _error "Cannot apply ci_presets '$name': key does not exist :/"
    fi
```
please let me know if this preset `icelake_notebook_gating_test` was actually triggered

> As for the prom metrics, looks like only_initialize=True for prom.register is not doing what it's supposed to do.. Yay

indeed, this initialization must not be triggered at all, that's why no metrics is registered
please try to find where this is going wrong, and please add a check which logs an error if the metrics array is empty
and catch the error with something like this:
```
        if grep "^ERROR" "$VISU_LOG_FILE"; then
            _warning "An error happened during the report generation :/."
            grep "^ERROR" "$VISU_LOG_FILE" > "$ARTIFACT_DIR"/FAILURE
            retcode=1
        fi
```
I don't like at 100% this mechanism of parsing a log file to detect errors, but at the same time it's a nice way to catch *minor* errors for which we don't want to abort the execution, but still notify that something went wrong.

maybe update the block above to `grep` in all the `matbench_*.log` files

kpouget:
@kdvalin I rebase the PR to integrate the latest changes

testing/notebooks/scale_test.sh
```diff
@@ -211,6 +211,8 @@ main() {
             return $failed
             ;;
         "run_tests_and_plots")
+            connect_ci
+
```
kpouget, line 215: @kdvalin can you remove this commit with the workaround?

kpouget:
thanks @kdvalin , I rebase and merge

PR #754, opened 2023-06-07
WIP: Ansible LLM Schema Models 

kdvalin:
This PR introduces Pydantic models for Ansible LLM testing

### DIFF
--- a/subprojects/matrix-benchmarking-workloads/ansible-llm/models/calculated.py
+++ b/subprojects/matrix-benchmarking-workloads/ansible-llm/models/calculated.py
@@ -1,0 +1,14 @@
+from matrix_benchmarking import models as matbench_models
+
+class StatusCodeDistribution(matbench_models.ExclusiveModel):
+    OK: int
+    # Don't know what other status codes there are
+
+class LatencyDistribution(matbench_models.ExclusiveModel):
+    percentage: int
+    latency: int
+
+class HistogramData(matbench_models.ExclusiveModel):
+    mark: float
+    count: int
+    frequency: float
--- a/subprojects/matrix-benchmarking-workloads/ansible-llm/store.py
+++ b/subprojects/matrix-benchmarking-workloads/ansible-llm/store.py
@@ -1,0 +1,6 @@
+from . import models
+
+import matrix_benchmarking.store as store
+
+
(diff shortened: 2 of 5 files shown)


kdvalin:
Would want sign off from @dagrayvid before merging

subprojects/matrix-benchmarking-workloads/ansible-llm/store.py
```diff
@@ -0,0 +1,6 @@
+from . import models
+
+import matrix_benchmarking.store as store
+
+
+store.register_custom_schema(models.AnsibleLLMPayload)
```
kpouget, line 6: I changed this to `register_lts_schema`

subprojects/matrix-benchmarking-workloads/ansible-llm/models/calculated.py
```diff
@@ -0,0 +1,14 @@
+from matrix_benchmarking import models as matbench_models
+
+class StatusCodeDistribution(matbench_models.ExclusiveModel):
+    OK: int
+    # Don't know what other status codes there are
```
dagrayvid, line 5:
```suggestion
class StatusCodeData(matbench_models.ExclusiveModel):
    status: str
    count: int
```

We may want to keep this more general, and then use a List in the AnsibleLLMData class, similar to LatencyDistribution.  There are many potential error statuses and I'm not sure if there is a way to enumerate all of them in maintainable way, in case e.g. the runtime or modelmesh adds new error codes / messages

dagrayvid: Hmmm, well we will need to do some conversion to turn the .json output from the load tester into this format anyway, so we can likely make this conversion as part of that.

kdvalin: We can still do the format below, just might not be as nice to deal with (and not use the code committed above)
```
{
   "OK": 1024,
   "err1": 2,
   ...
}
```

kpouget:
I close this PR as we're deprecating `openshift-psap/ci-artifacts` in favor of `openshift-psap/topsail`. Please reopen the PR there if relevant.

PR #768, opened 2023-06-09
load-aware: add ansible role to deploy Trimaran scheduler

drewrip:
This should cover the basics of deploying the Trimaran scheduler.

Trimaran can be deployed to enable load aware scheduling using:

```
./run_toolbox.py load_aware deploy_trimaran
```

### DIFF
--- a/roles/load_aware_deploy_trimaran/files/trimaran.yaml
+++ b/roles/load_aware_deploy_trimaran/files/trimaran.yaml
@@ -5,11 +5,11 @@ metadata:
   name: trimaran
 ---
 apiVersion: v1
-kind: ConfigMap
+kind: Secret
 metadata:
   name: trimaran-scheduler-config
   namespace: trimaran
-data:
+stringData:
   config.yaml: |
     apiVersion: kubescheduler.config.k8s.io/v1beta2
     kind: KubeSchedulerConfiguration
@@ -75,8 +75,8 @@ spec:
       serviceAccount: trimaran-scheduler
       volumes:
       - name: etckubernetes
-        configMap:
-          name: trimaran-scheduler-config
+        secret:
+          secretName: trimaran-scheduler-config
       containers:
         - name: trimaran-scheduler
           env:
--- a/roles/load_aware_deploy_trimaran/files/trimaran.yaml
+++ b/roles/load_aware_deploy_trimaran/files/trimaran.yaml
@@ -1,0 +1,93 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: trimaran
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: trimaran-scheduler-config
+  namespace: trimaran
+data:
+  config.yaml: |
+    apiVersion: kubescheduler.config.k8s.io/v1beta2
+    kind: KubeSchedulerConfiguration
+    leaderElection:
+      leaderElect: false
+    profiles:
+    - schedulerName: trimaran-scheduler
+      plugins:
+        score:
+          disabled:
+          - name: NodeResourcesBalancedAllocation
+          - name: NodeResourcesLeastAllocated
+          enabled:
+          - name: TargetLoadPacking
+      pluginConfig:
+      - name: TargetLoadPacking
+        args:
+          defaultRequests:
+            cpu: "2000m"
+          defaultRequestsMultiplier: "1"
+          targetUtilization: 1
+          metricProvider:
+            type: Prometheus
+            address: https://${THANOS_MONITORING_ENDPOINT}
+            token: $MONITORING_TOKEN
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: trimaran-scheduler
+  namespace: trimaran
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: trimaran
+subjects:
+- kind: ServiceAccount
+  name: trimaran-scheduler
+  namespace: trimaran
+roleRef:
+  kind: ClusterRole
+  name: system:kube-scheduler
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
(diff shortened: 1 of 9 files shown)


drewrip:
[test passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/768/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1668223260929560576)

kpouget:
looks good @drewrip, merging.
Some hints for follow-up PRs:
- add a matching `undeploy` role, so that we can use this in the bare-metal cluster
- handle the `Secret` independently of the other resources, so that you can make a copy of the other resources to a `src` directory (it's nice to have an quick way to see the resources created by the role, from the artifacts)
- capture some state artifacts to "confirm" that everything went well during the deployment. Can be at least: `oc get all -n trimaran > .../all.status`
- I don't know yet know to confirm that Trimaran is used as a scheduler, but maybe: create a dummy Pod, let it be scheduled and terminate, verify in its spec that Trimaran was used as scheduler instead of the default one

I'm curious about this last one actually, we need an answer for it. Does trimaran sets up itself as default scheduler, or do you have to specify in the pod spec that it should be used?

roles/load_aware_deploy_trimaran/files/trimaran.yaml
```diff
@@ -0,0 +1,93 @@
...
+        - name: trimaran-scheduler
+          env:
+          - name: ENABLE_OPENSHIFT_AUTH
+            value: "true"
+          image: quay.io/chenw615/kube-scheduler:ocp
+          imagePullPolicy: Always
+          args:
+          - /bin/kube-scheduler
+          - --config=/etc/kubernetes/config.yaml
+          - -v=6
```
kpouget, line 90: another possible update: allow the configuration of the verbosity level from the Python entrypoint

testing/load-aware/test.py
```diff
@@ -65,7 +65,7 @@ def prepare_ci():
     install_ocp_pipelines()
 
     run.run("./run_toolbox.py from_config cluster capture_environment --suffix sample")
-
+    run.run("./run_toolbox.py load_aware deploy_trimaran")
```
kpouget, line 68: another possible update: add a configuration flag to `testing/pipelines/config.yaml` with: `deploy_trimaran: false`, and if set to true (either from the PR config or from the `load-aware` preset), call this command

kpouget:
[test results](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/768/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1668223260929560576/artifacts/light/test/artifacts/001__plots/report_00_report:_error_report.html) and [gather-extra](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/768/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1668223260929560576/artifacts/light/gather-extra/artifacts/)

roles/load_aware_deploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    set -o pipefail;
+    set -o nounset;
+    set -o errtrace;
+
+    export MONITORING_SECRET=$(oc get secret -n openshift-user-workload-monitoring
+      | grep  prometheus-user-workload-token
+      | head -n 1
+      | awk '{print $1 }'
+    )
+
```
kpouget, line 29: @drewrip can you add here:
```
if [[ -z "$MONITORING_SECRET" ]]; then
  echo "Coudln't not get the prometheus-user-workload-token secret, aborting"
  exit 1
fi
```

roles/load_aware_deploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+    export MONITORING_TOKEN=$(echo $(oc get secret $MONITORING_SECRET -n openshift-user-workload-monitoring -o json | jq -r '.data.token')
```
kpouget, line 30:
```
export MONITORING_TOKEN=$(oc get secret $MONITORING_SECRET_NAME -n openshift-user-workload-monitoring -o json | jq -r .data.token | base64 -d)
```
(no need for the nest `$()`)

roles/load_aware_deploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+      | base64 -d)
+
+    export THANOS_MONITORING_ENDPOINT=$(oc get route thanos-querier -n openshift-monitoring -o json
+      | jq -r '.spec.host')
+
```
kpouget, line 35: echo "Route to thanos monitoring endpoint: $THANOS_MONITORING_ENDPOINT"

roles/load_aware_deploy_trimaran/tasks/main.yml
drewrip: Yes! Can do. Will include this in a PR with improvements.

roles/load_aware_deploy_trimaran/tasks/main.yml
drewrip: I'll also include this in a PR with general fixes/improvements.

I also found this a bit odd, it actually came from a portion of the OpenShift docs:

> Procedure
> Extract a token to connect to Prometheus:
> ```
> $ TOKEN=`echo $(oc get secret $SECRET -n openshift-user-workload-monitoring -o json | jq -r '.data.token') | base64 -d`
> ```

https://docs.openshift.com/container-platform/4.10/monitoring/enabling-monitoring-for-user-defined-projects.html#enabling-monitoring-for-user-defined-projects_enabling-monitoring-for-user-defined-projects

testing/load-aware/test.py
drewrip: Also will be included in improvements PR!

PR #780, opened 2023-06-13
WIP: role to undeploy trimaran

drewrip:


roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+- name: Delete Trimaran namespace
+  shell:
+    oc delete ns trimaran
+  
+- name: Delete user monitoring namespace
+  shell:
+    oc delete ns openshift-user-workload-monitoring
```
kpouget, line 7: you cannot delete this namespace, we don't know if we created it
let this part (the activation of the user-workload-monitoring) untouched, it's too complex to handle properly

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+- name: Delete Trimaran namespace
+  shell:
+    oc delete ns trimaran
```
kpouget, line 3: this role is quite fine (except the task below),
but what I would prefer is:
in `roles/load_aware_undeploy_trimaran/files/`, create a symlink to `roles/load_aware_deploy_trimaran/files/trimaran.yaml`
and in this file, just `oc delete -f {{ file }}`
this way, you know that no matter what you add to the file above, it will be properly deleted

at least I hope it will work, as the namespace is first in the yaml file 🤔

roles/load_aware_undeploy_trimaran/tasks/main.yml
drewrip: You're right. I meant to delete the config map that we purposefully create to setup user application monitoring

https://github.com/openshift-psap/ci-artifacts/blob/main/roles/load_aware_deploy_trimaran/files/cluster-monitoring-config.yaml

roles/load_aware_undeploy_trimaran/tasks/main.yml
drewrip: This feels much safer. I adopted this in the most recent commit! I have a variable in vars/main/resources.yml pointing to the deploy_trimaran configuration files rather than a symlink though. Not sure if there was a reason for one over the other.

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+- name: Delete Trimaran resources
+  shell: |
+    oc delete -f {{ trimaran_setup_config }} 2>&1
```
kpouget, line 3: no need for the `|`
let me explain how this works, it's not straightforward ;-)
```
  shell: |
    echo hello
    echo world
    echo hello \
              world
```
this is a *multi-line* string. No need for  `;`. This works and outputs:
```
hello
world
hello world
```
  shell:
    echo hello
    echo world
    echo hello \
              world
this is a *single-line* string. `;` required, as all the "lines" above are concatenated. This outputs:
```
hello echo world echo hello world
```
(not 100% sure about the last `/`, just don't do it ;-))

kpouget: can you also remove the `2>&1` and turn `shell` into `command`?

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+- name: Delete Trimaran resources
```
kpouget, line 1: can you check first if the `trimaran` namespace exists?
```
command: oc get ns/trimaran --ignore-not-found -oname
register: has_trimaran
```
and then:
```
command: oc delete ...
when: has_trimaran.stdout | length > 0
``

File: roles/load_aware_undeploy_trimaran/tasks/main.yml [hunk 5]
```diff
@@ -0,0 +1,21 @@
...
+  shell: |
+    oc delete -f {{ trimaran_setup_config }} 2>&1
+  register: delete_trimaran
+  failed_when: delete_trimaran.rc != 0 and 'namespaces "trimaran" not found' not in delete_trimaran.stdout
+
+- name: Delete user monitoring configmap
+  shell: |
+    oc delete -f {{ user_applications_monitor_config }} 2>&1
+  register: delete_monitoring
+  failed_when: delete_monitoring.rc != 0 and 'configmaps "cluster-monitoring-config" not found' not in delete_monitoring.stdout
```
kpouget, line 11: I was going to suggest this:
```suggestion
  command:
    oc delete -f {{ user_applications_monitor_config }} --ignore-not-found
```
but I prefer that we don't touch this config map, we don't know who created it ... 
(and we don't really mind if user workload monitoring remains enabled)

roles/load_aware_undeploy_trimaran/tasks/main.yml
kpouget: > I have a variable in vars/main/resources.yml pointing to the deploy_trimaran configuration files rather than a symlink though. 

yes, works equally. One over the other is mostly a matter of taste,
my idea is to keep the role "standalone", so at the Ansible level, it's not referencing anything outside of the role. 
but at the FS level it is, so the encapsulation isn't complete anyway ....

roles/load_aware_undeploy_trimaran/tasks/main.yml
drewrip: I can go ahead and just remove this, but to deploy Trimaran we do have to first deploy this config map.
https://github.com/openshift-psap/ci-artifacts/blob/4b7be770286039f0bc03c254d31c43e6d66d3474/roles/load_aware_deploy_trimaran/tasks/main.yml#LL1C1-L3C55

As you mentioned it's probably not a big deal if it _does_ stay configured, but we are actually the ones that create it :)

drewrip:
@kpouget  The latest commit should resolve your concerns, pending how to proceed with https://github.com/openshift-psap/ci-artifacts/pull/780#discussion_r1229539085

roles/load_aware_undeploy_trimaran/tasks/main.yml
kpouget: yes, I prefer that we don't touch this resource (=we just create it in the deployment stage)
rational is:
- if nobody enabled it before, we enable it and let it enabled
- if somebody enabled it before, we cannot (easily) know it. So just in case, we let it enabled

kpouget: just FYI, the reason why it's not enabled by default is that cluster admins don't want that _any_ payload installed by the cluster users to generate metrics that are stored in Prometheus.
In a controlled environment where we trust the workload that we deploy, it's safe to let this feature enabled

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  when: has_trimaran.stdout | length > 0
+
+- name: Ensure user applications monitor config is deployed
+  command: oc get configmap cluster-monitoring-config --ignore-not-found -oname -n openshift-monitoring
+  register: has_user_monitoring
+
+- name: Delete user monitoring configmap
+  command:
+    oc delete -f {{ user_applications_monitor_config }} --ignore-not-found
+  when: has_user_monitoring | length > 0
```
kpouget, line 17: this should go away, as discussed in the other thread

just FYI, `when: has_user_monitoring | length > 0` this isn't necessary, as `--ignore-not-found` prevents `oc delete` from crashing (complaining) if the ConfigMap does not exist

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  retries: 30
+
+- name: Wait for monitoring configmap to disappear
+  shell: |
+    oc get configmap cluster-monitoring-config --no-headers 2> /dev/null
+  register: has_user_monitoring_deleted
+  until: has_user_monitoring_deleted.rc != 0
+  failed_when: has_user_monitoring_deleted.rc == 0
+  delay: 5
+  retries: 30
```
kpouget, line 35: this isn't necessary, when you do
```
oc delete <resource>
```
the command will block until the resource is destroyed (except is you pass `--wait=false`)

the same applied to the namespace deletion above actually

roles/load_aware_undeploy_trimaran/tasks/main.yml
drewrip: Makes sense!

drewrip:
[d0caa25](https://github.com/openshift-psap/ci-artifacts/pull/780/commits/d0caa25027411d5b02fe258c33fdb431176f23c5)No longer deletes the ConfigMap and removes the redundancy that was pointed out by @kpouget.

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/780/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1668988997793222656/artifacts/light/test/artifacts/002__loadaware__undeploy_trimaran/_ansible.log) Trimaran was successfully undeployed.

[All test artifacts](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/780/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1668988997793222656/)

testing/load-aware/test.py
```diff
@@ -116,6 +116,8 @@ def test_ci():
         else:
             logging.warning("Not generating the visualization as the test artifact directory hasn't been created.")
 
+    # Temporary cleanup test for Trimaran
+    cleanup_cluster()
```
kpouget, line 120:
```suggestion
```

kpouget:
thanks @drewrip , I've removed the workaround to trigger the undeployment, and I merge the PR :)

PR #781, opened 2023-06-13
Update testing/model-mesh scripts

dagrayvid:


kpouget:
@dagrayvid I think only the visualization step failed, 
that's a stub that I put in place quite a while ago, that was never used. I guess you can just cut this part from the test execution to avoid failing.

dagrayvid:
@kpouget ack, though I think more needs to be done here. I should switch 1-install-mm.sh to install downstream RHODS.

kpouget:
> @kpouget ack, though I think more needs to be done here. I should switch 1-install-mm.sh to install downstream RHODS.

sure, but making a test turn from 🔴 to 🟢 with one line is always good to take :D 
```
git rm testing/model-mesh/7-analyze.sh
```
Rational behind it is: its always easier when we jump from 🟢 to 🟢, but clearly here more work will be required. 
Maybe a first client for the `skeleton` template? (The `skeleton` guide will come [with its own little legs] when we/I will start using it to integrate a new workload)

kpouget:
I close this PR as we're deprecating `openshift-psap/ci-artifacts` in favor of `openshift-psap/topsail`. Please reopen the PR there if relevant.

PR #796, opened 2023-06-20
load-aware: enable configuring Trimaran plugins and settings

drewrip:
Adds the ability to configure Trimaran's plugin and corresponding settings from the CI. Also adds the ability to change the log level. This PR should not affect the `undeploy` role. However, the `undeploy` role must have access to the generated `trimaran.yaml` that was used to deploy Trimaran or it might not be able to properly cleanup.

### DIFF
--- a/roles/load_aware_deploy_trimaran/.gitignore
+++ b/roles/load_aware_deploy_trimaran/.gitignore
@@ -1,1 +1,0 @@
-files/trimaran.yaml
--- a/roles/load_aware_deploy_trimaran/.gitignore
+++ b/roles/load_aware_deploy_trimaran/.gitignore
@@ -1,0 +1,1 @@
+files/trimaran.yaml
--- a/roles/load_aware_deploy_trimaran/tasks/main.yml
+++ b/roles/load_aware_deploy_trimaran/tasks/main.yml
@@ -47,6 +47,16 @@
   retries: 30
   until: token_size.stdout | int > 1000 
 
+- name: Create the src artifacts directory
+  file:
+    path: "{{ artifact_extra_logs_dir }}/src/"
+    state: directory
+    mode: '0755'
+
+- name: Define location of Trimaran config
+  set_fact:
+    trimaran_setup_config: "{{ artifact_extra_logs_dir }}/src/trimaran-templated-config.yaml"
+
 - name: Configure Trimaran options
   template:
     src: "{{ trimaran_setup_template }}"
@@ -93,4 +103,3 @@
     oc get pod trimaran-test -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.yaml"
     oc describe pod trimaran-test -n trimaran > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.desc"
     oc get all -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/all-trimaran.yaml"
-    cp {{ trimaran_setup_config }} "{{ artifact_extra_logs_dir }}/trimaran-templated-config.yaml"
--- a/roles/load_aware_deploy_trimaran/tasks/main.yml
+++ b/roles/load_aware_deploy_trimaran/tasks/main.yml
@@ -47,14 +47,18 @@
   retries: 30
   until: token_size.stdout | int > 1000 
 
+- name: Configure Trimaran options
+  template:
+    src: "{{ trimaran_setup_template }}"
+    dest: "{{ trimaran_setup_config }}"
+
 - name: Deploy Trimaran scheduler
   shell: | 
     set -o errexit;
     set -o pipefail;
     set -o nounset;
     set -o errtrace;
-   
-    export THANOS_MONITORING_ENDPOINT={{ thanos_endpoint.stdout }}
+
     echo "Route to thanos monitoring endpoint: {{ thanos_endpoint.stdout }}"
     export MONITORING_TOKEN=$(oc get secret {{ monitoring_secret.stdout }} -n openshift-user-workload-monitoring -o json | jq -r '.data.token' | base64 -d)
     cat {{ trimaran_setup_config }} | envsubst | oc apply -f -
@@ -89,7 +93,4 @@
     oc get pod trimaran-test -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.yaml"
     oc describe pod trimaran-test -n trimaran > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.desc"
     oc get all -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/all-trimaran.yaml"
-
-- name: Test
-  shell: |
-    echo "Using plugin: {{ load_aware_deploy_trimaran_plugin }}"
+    cp {{ trimaran_setup_config }} "{{ artifact_extra_logs_dir }}/trimaran-templated-config.yaml"
--- a/roles/load_aware_deploy_trimaran/tasks/main.yml
+++ b/roles/load_aware_deploy_trimaran/tasks/main.yml
@@ -55,8 +55,8 @@
     set -o errtrace;
    
     export THANOS_MONITORING_ENDPOINT={{ thanos_endpoint.stdout }}
-    export MONITORING_SECRET={{ monitoring_secret.stdout }}
-    export MONITORING_TOKEN=$(oc get secret $MONITORING_SECRET -n openshift-user-workload-monitoring -o json | jq -r '.data.token' | base64 -d)
+    echo "Route to thanos monitoring endpoint: {{ thanos_endpoint.stdout }}"
+    export MONITORING_TOKEN=$(oc get secret {{ monitoring_secret.stdout }} -n openshift-user-workload-monitoring -o json | jq -r '.data.token' | base64 -d)
     cat {{ trimaran_setup_config }} | envsubst | oc apply -f -
 
   register: deploy_trimaran
@@ -84,7 +84,12 @@
   retries: 20
   until: trimaran_test_pod_state.stdout == 'Completed'
 
-- name: Dump trimaran test pod info
+- name: Dump trimaran info
   shell: |
     oc get pod trimaran-test -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.yaml"
     oc describe pod trimaran-test -n trimaran > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.desc"
+    oc get all -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/all-trimaran.yaml"
+
+- name: Test
+  shell: |
+    echo "Using plugin: {{ load_aware_deploy_trimaran_plugin }}"
--- a/roles/load_aware_deploy_trimaran/tasks/main.yml
+++ b/roles/load_aware_deploy_trimaran/tasks/main.yml
@@ -56,18 +56,20 @@
    
     export THANOS_MONITORING_ENDPOINT={{ thanos_endpoint.stdout }}
     export MONITORING_SECRET={{ monitoring_secret.stdout }}
-    export MONITORING_TOKEN=$(echo $(oc get secret $MONITORING_SECRET -n openshift-user-workload-monitoring -o json | jq -r '.data.token') | base64 -d)
+    export MONITORING_TOKEN=$(oc get secret $MONITORING_SECRET -n openshift-user-workload-monitoring -o json | jq -r '.data.token' | base64 -d)
     cat {{ trimaran_setup_config }} | envsubst | oc apply -f -
 
   register: deploy_trimaran
 
 - name: Ensure Trimaran is Running
   shell:
-    oc get pods -n trimaran | awk 'NR > 1 {print $3}'
+    oc get pods -n trimaran
+      | grep "trimaran-scheduler"
+      | awk '{print $3}'
   register: trimaran_running
   delay: 3
   retries: 20
-  until: "trimaran_running.stdout == 'Running'"
+  until: trimaran_running.stdout == 'Running'
 
 - name: Deploy test pod with Trimaran
   shell:
@@ -81,3 +83,8 @@
   delay: 5
   retries: 20
   until: trimaran_test_pod_state.stdout == 'Completed'
+
+- name: Dump trimaran test pod info
+  shell: |
+    oc get pod trimaran-test -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.yaml"
+    oc describe pod trimaran-test -n trimaran > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.desc"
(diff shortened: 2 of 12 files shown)


roles/load_aware_deploy_trimaran/.gitignore
```diff
@@ -0,0 +1 @@
+files/trimaran.yaml
```
kpouget, line 1: please remove this

roles/load_aware_deploy_trimaran/tasks/main.yml
```diff
@@ -47,27 +47,33 @@
   retries: 30
   until: token_size.stdout | int > 1000 
 
+- name: Configure Trimaran options
+  template:
+    src: "{{ trimaran_setup_template }}"
+    dest: "{{ trimaran_setup_config }}"
+
```
kpouget, line 54: please add here something like this:
```
- name: Create the src artifacts directory
  file:
    path: "{{ artifact_extra_logs_dir }}/src/"
    state: directory
    mode: '0755'

- name: define the variables
  set_fact:
    trimaran_setup_config: "{{ artifact_extra_logs_dir }}/src/trimaran-templated-config.yaml"

- name: Configure Trimaran options
  template:
    src: "{{ trimaran_setup_template }}"
    dest: "{{ trimaran_setup_config }}"
```
(the `set_fact` task isn't mandatory, it's just to define a variable/constant instead of having a harcoded path duplicated here and there. In theory that's preferable, but in practice I often omit it :#)

roles/load_aware_deploy_trimaran/tasks/main.yml
```diff
@@ -81,3 +87,10 @@
   delay: 5
   retries: 20
   until: trimaran_test_pod_state.stdout == 'Completed'
+
+- name: Dump trimaran info
+  shell: |
+    oc get pod trimaran-test -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.yaml"
+    oc describe pod trimaran-test -n trimaran > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.desc"
+    oc get all -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/all-trimaran.yaml"
+    cp {{ trimaran_setup_config }} "{{ artifact_extra_logs_dir }}/trimaran-templated-config.yaml"
```
kpouget, line 96: this can be removed after the suggestions above have been applied

roles/load_aware_deploy_trimaran/vars/main/resources.yml
```diff
@@ -1,3 +1,4 @@
 user_applications_monitor_config: "roles/load_aware_deploy_trimaran/files/cluster-monitoring-config.yaml"
+trimaran_setup_template: "roles/load_aware_deploy_trimaran/templates/trimaran.yaml.j2"
 trimaran_setup_config: "roles/load_aware_deploy_trimaran/files/trimaran.yaml"
```
kpouget, line 3: this line can be removed

testing/load-aware/config.yaml
```diff
@@ -96,3 +102,11 @@ matbench:
   ignore_exit_code: true
   # directory to plot. Set by testing/common/visualize.py before launching the visualization
   test_directory: null
+load_aware:
```
kpouget, line 105: can you move this block before `matbench`? it's just out of consistency with the other config files. The order is: cluster settings, RHODS settings if applicable, test settings, visualization settings, you get the idea

toolbox/load_aware.py
```diff
@@ -10,14 +10,41 @@ class LoadAware:
 
     @AnsibleRole("load_aware_deploy_trimaran")
     @AnsibleMappedParams
-    def deploy_trimaran(self):
+    def deploy_trimaran(self, log_level=1,
```
kpouget, line 13: I think you didn't regenerate the role's config file:
```
./run_toolbox.py repo generate_ansible_default_settings
```

kpouget (commented):
thanks @drewrip, looks very good :) I just put a bunch of comments to have the template file instanciated directly in the `{{ artifact_extra_logs_dir }}/src` directory, but I'm happy you got the workflow correct \o/

drewrip:
@kpouget thanks for the suggestions! Glad to hear the workflow looks correct. I'll make some fixes per your suggestions.

roles/load_aware_deploy_trimaran/tasks/main.yml
drewrip: Do we no longer want to save this info or does it automatically get saved with the new path?

kpouget: we do save it :)
here is the way I would write the block above, which makes it clearer that it's stored in the artifacts:
```
- name: Create the src artifacts directory
  file:
    path: "{{ artifact_extra_logs_dir }}/src/"
    state: directory
    mode: '0755'

- name: Configure Trimaran options
  template:
    src: "{{ trimaran_setup_template }}"
    dest: "{{ artifact_extra_logs_dir }}/src/trimaran-resources.yaml"
```

drewrip:
@kpouget Latest commit should implement most of the suggestions. Moving the location of `trimaran.yaml` breaks `undeploy` as the location of the configuration that deploys Trimaran is no longer in a predictable location as far as I can tell. My fix just added a default value for each value in the template and runs through the template again with default values when undeploying. The specific settings we are configuring don't matter when deleting anyway.

Also, I didn't fully remove the lines
```
oc get pod trimaran-test -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.yaml"
oc describe pod trimaran-test -n trimaran > "{{ artifact_extra_logs_dir }}/trimaran-test.pod.desc"
oc get all -n trimaran -oyaml > "{{ artifact_extra_logs_dir }}/all-trimaran.yaml"
cp {{ trimaran_setup_config }} "{{ artifact_extra_logs_dir }}/trimaran-templated-config.yaml"
```

I just removed the last, as we wanted to capture the first 3, and I couldn't see how they would get saved after removing this task.

kpouget:
> I just removed the last, as we wanted to capture the first 3, and I couldn't see how they would get saved after removing this task.

yes, sure, I see that my comment is confusing, but of course I meant to remove only the last line 😀 
```
cp {{ trimaran_setup_config }} "{{ artifact_extra_logs_dir }}/trimaran-templated-config.yaml"
```

toolbox/load_aware.py
```diff
@@ -10,14 +10,41 @@ class LoadAware:
...
+                        plugin="TargetLoadPacking", default_requests_cpu="2000m",
+                        default_target_requests_multiplier="2", target_utilization=70,
```
kpouget, line 15: > default_target_requests_multiplier="2"
why is this one a `str`, not an `int`? 🤔 
and can you put all the parameters in a dedicated line? (just for consistency), and with a comma after the last one

roles/load_aware_deploy_trimaran/templates/trimaran.yaml.j2
```diff
@@ -23,17 +23,22 @@ stringData:
           - name: NodeResourcesBalancedAllocation
           - name: NodeResourcesLeastAllocated
           enabled:
-          - name: TargetLoadPacking
+          - name: {{ load_aware_deploy_trimaran_plugin | default('NA') }}
```
kpouget, line 26: I think you don't need `| default('NA')`, the variables will always have a value. Or fail earlier if they must have a value and the user didn't provide one

kpouget: hum, or you pass the defaults for the `undeploy` role, is that it? makes sense then

kpouget: not sure what's the proper way to go 🤔 🤔 🤔 
I don't like these `| default('NA')` in the template, they complicate the reading with unnecessary things.

🤔 
we need to find a way to pass all the mandatory variable in the `undeploy` function ...

drewrip: The `| default('NA')` definitely feels hacky. Let me see if I can think of something better that doesn't clog up the template.

toolbox/load_aware.py
drewrip: Now that is a good question! That was actually intentional. For some reason that field is a string in the config

https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/trimaran/targetloadpacking

Not sure why though, will have to make sure their example is correct.

kpouget: ack, if it's on purpose, keep it as you think is most suitable!

drewrip:
@kpouget This should fix it. I implemented this before seeing your comment, but it should keep them both in sync without having to maintain them both. It was still missing the `thanos_endpoint` so there is a dummy command to set one, but I believe both solutions would require something like this.

I don't think there will be many more settings or configuration to add so I can just manually set the options too as you pointed out, probably wouldn't be a big deal.

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -1,3 +1,25 @@
...
+- name: Define the Trimaran setup config
+  set_fact:
+    trimaran_setup_config: "{{ artifact_extra_logs_dir }}/src/trimaran-templated-config.yaml"
+
+- name: Define the same default settings from deploy role
+  include_vars: "{{ trimaran_deploy_defaults }}"
+
+- name: Dummy Thanos Endpoint
+  command: echo "NA"
+  register: thanos_endpoint
```
kpouget, line 16: this construct is required because you have this in the template:
```
address: https://{{ thanos_endpoint.stdout }}
```

I prefer doing it this way (in the deploy code), but your solution works as well:
```
- name: Get Thanos Endpoint
  shell:
    oc get route thanos-querier -n openshift-monitoring -o json
      | jq -r '.spec.host'
register: thanos_endpoint_cmd

- name: set thanos endpoint variable
  set_fact:
    thanos_endpoint: " https://{{ thanos_endpoint_cmd.stdout }}"
```
and then in the template, keep it simple:
```
address: {{ thanos_endpoint }}
```

kpouget (commented):
last thing, regarding the execution of the cleanup
look how I'm doing it there: https://github.com/openshift-psap/ci-artifacts/blob/main/testing/pipelines/test.py#L440
```
        if config.ci_artifacts.get_config("clusters.cleanup_on_exit"):
            pipelines_cleanup_cluster()
```
this way, I can launch the test with
```
/test ...
/var clusters.cleanup_on_exit=true
```
to force the execution of the cleanup path as part of the CI validation run.
Makes it easier to troubleshoot the cleanup step :D

drewrip:
@kpouget Added the flag to trigger cleanup. Let me know if this method of inheriting the variables from deploy looks alright!

drewrip:
[test passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/796/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1671480360149454848). Should be good to merge.

The [consistency check](https://github.com/openshift-psap/ci-artifacts/actions/runs/5333335953/jobs/9663671620?pr=796) failed, as it complains the `thanos_endpoint` variable wasn't used in the undeploy role. It is used in the undeploy role, when templating the dummy config.

roles/load_aware_undeploy_trimaran/vars/main/resources.yml
```diff
@@ -1 +1,3 @@
-trimaran_setup_config: "roles/load_aware_deploy_trimaran/files/trimaran.yaml"
+trimaran_setup_template: "roles/load_aware_deploy_trimaran/templates/trimaran.yaml.j2"
+trimaran_deploy_defaults: "roles/load_aware_deploy_trimaran/defaults/main/config.yml"
+thanos_endpoint: "NA"
```
kpouget, line 3: need to remove this

roles/load_aware_undeploy_trimaran/tasks/main.yml
```diff
@@ -1,3 +1,21 @@
+- name: Create the src artifacts directory
+  file:
+    path: "{{ artifact_extra_logs_dir }}/src/"
+    state: directory
+    mode: '0755'
+
+- name: Define the Trimaran setup config
+  set_fact:
+    trimaran_setup_config: "{{ artifact_extra_logs_dir }}/src/trimaran-templated-config.yaml"
+
```
kpouget, line 10: here, define
```
 thanos_endpoint: "NA"
```

kpouget:
> [test passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/796/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1671480360149454848). 

where you share the test link, please pass the link to _the most relevant_ artifact file ;-) For this test, it can be:
* [loadaware__deploy_trimaran ran with the default configuration](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/796/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1671480360149454848/artifacts/light/prepare/artifacts/002__loadaware__deploy_trimaran/)
* [002__loadaware__undeploy_trimaran role ran](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/796/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1671480360149454848/artifacts/light/test/artifacts/002__loadaware__undeploy_trimaran/)

> The [consistency check](https://github.com/openshift-psap/ci-artifacts/actions/runs/5333335953/jobs/9663671620?pr=796) failed, as it complains the thanos_endpoint variable wasn't used in the undeploy role. It is used in the undeploy role, when templating the dummy config.
we need to make it pass :)
I added two comments to remove the offensive line, and add the definition in a `set_fact` task

kpouget:
thanks @drewrip , testing and merging
/test load-aware-light
/var clusters.cleanup_on_exit=true

kpouget:
arf
```
PR_POSITIONAL_ARGS='Should be good to merge.'
PR_POSITIONAL_ARG_1='Should'
PR_POSITIONAL_ARG_2='be'
PR_POSITIONAL_ARG_3='good'
PR_POSITIONAL_ARG_4='to'
PR_POSITIONAL_ARG_5='merge.'
PR_POSITIONAL_ARG_0='load-aware-light'
```
I should improve the code doing the [PR parsing](https://github.com/openshift-psap/ci-artifacts/blob/main/testing/utils/openshift_ci/pr_args.sh) (should actually turn it into Python), there's something wrong with it :D

drewrip:
tests passed.

[deploying test](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/796/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1671543665677307904/artifacts/light/prepare/artifacts/002__loadaware__deploy_trimaran/_ansible.log)
[undeploying test](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/796/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1671543665677307904/artifacts/light/test/artifacts/002__loadaware__undeploy_trimaran/_ansible.log)

PR #801, opened 2023-06-22
pipelines: sample pipeline with multiple dependent stages and fixed runtime

drewrip:
It was brought up that this pipeline might be useful for testing. It has 4 parallel pipelines, each 3 interdependent stages. Each stage should take 2 minutes and generate CPU load during that time by repeatedly taking square roots.

### DIFF
--- a/testing/pipelines/config.yaml
+++ b/testing/pipelines/config.yaml
@@ -51,6 +51,9 @@ ci_presets:
     tests.pipelines.notebook_filename: kfp_hash.ipynb
     tests.pipelines.notebook_directory: testing/pipelines/notebooks/hash
 
+  pipeline_timed_multistage:
+    tests.pipelines.notebook_filename: kfp_timed_multistage.ipynb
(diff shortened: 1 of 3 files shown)


drewrip:
@kpouget I've tried to get this working in the past and it isn't straight forward. Tekton does some of its own parsing so it can clearly tell which stages depend on which and how to construct the DAG. So from what I can tell you are a bit limited as to what type of control flow you use when creating the pipeline. Tekton didn't like compiling `for` loops or putting the results of pipelines in arrays.

I can look into it and see if Tekton has a way to officially support this though!

kpouget:
ack, I'm not surprised!
there's one way do to what I suggested, it's called code generation :D 
write what you want to do in a Python file, and get it executed ;-) But that might be going a bit too far, as we're not even sure to be able to use Pipelines with the load-aware scheduler!

kpouget:
/lgtm
/approve
/hold 
@drewrip , you can merge it when you're ready with `/hold cancel`

PR #810, opened 2023-06-28
load-aware: role to launch load-aware scale test corresponding to Poisson distribution

drewrip:
This adds the role to launch load-aware scale tests according to a few different statistical distributions, specifically the Poisson. Options for gamma, uniform, normal, and a bimodal distributions are also included.

Currently this only launches test pods. The next step will be adding a real workload to the scale test.

### DIFF
--- a/roles/load_aware_scale_test/tasks/main.yml
+++ b/roles/load_aware_scale_test/tasks/main.yml
@@ -1,0 +1,63 @@
+- name: Generate load timeline
+  shell:
+    python3 {{ load_timeline_generator }} {{ load_aware_scale_test_distribution }} {{ load_aware_scale_test_duration}} {{ load_aware_scale_test_instances }} > "{{ artifact_extra_logs_dir }}/load_aware_load_timeline.txt"
+
+- name: Create namespace for default test
+  command:
+    oc apply -f {{ load_aware_ns }}
+
+- name: Run test workload with default scheduler and load timeline
+  shell:
+    cat "{{ artifact_extra_logs_dir }}/load_aware_load_timeline.txt"
+      | python3 {{ pod_start_scheduler }} default > "{{ artifact_extra_logs_dir }}/default_workload_scheduler.txt"
+
+- name: Wait for default workloads to finish
+  shell:
+    oc get pods -n load-aware | awk 'NR > 1 { print $3 }'
+  register: load_aware_workload
+  delay: 15
+  retries: 120
+  until:
+    "'Running' not in load_aware_workload.stdout
+      and 'Pending' not in load_aware_workload.stdout
+      and 'Failed' not in load_aware_workload.stdout
+      and 'ContainerCreating' not in load_aware_workload.stdout"
+
+- name: Dump info about default scheduler resources
+  shell:
+    oc get all -n load-aware -oyaml > "{{ artifact_extra_logs_dir }}/all_default_resources.yaml"
+
+- name: Cleanup namespace for default test
+  command:
+    oc delete -f {{ load_aware_ns }}
+
+- name: Create namespace for trimaran test
+  command:
+    oc apply -f {{ load_aware_ns }}
+
+
+- name: Run test pod with trimaran scheduler and load timeline
+  shell:
+    cat "{{ artifact_extra_logs_dir }}/load_aware_load_timeline.txt"
+      | python3 {{ pod_start_scheduler }} trimaran > "{{ artifact_extra_logs_dir }}/trimaran_workload_scheduler.txt"
+
+- name: Wait for default workloads to finish
+  shell:
+    oc get pods -n load-aware | awk 'NR > 1 { print $3 }'
+  register: load_aware_workload
+  delay: 15
+  retries: 120
+  until:
+    "'Running' not in load_aware_workload.stdout
+      and 'Pending' not in load_aware_workload.stdout
+      and 'Failed' not in load_aware_workload.stdout
+      and 'ContainerCreating' not in load_aware_workload.stdout"
+
+
+- name: Dump info about trimaran resources
+  shell:
+    oc get all -n load-aware -oyaml > "{{ artifact_extra_logs_dir }}/all_trimaran_resources.yaml"
+
+- name: Cleanup namespace for test
+  command:
+    oc delete -f {{ load_aware_ns }}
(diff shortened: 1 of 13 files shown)


drewrip:
fixed typo causing fail -> [log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1674075682960838656/build-log.txt)

roles/load_aware_scale_test/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+  retries: 120
+  until:
+    "'Running' not in load_aware_workload.stdout
+      and 'Pending' not in load_aware_workload.stdout
+      and 'Failed' not in load_aware_workload.stdout
+      and 'ContainerCreating' not in load_aware_workload.stdout"
+
+- name: Dump info about default scheduler resources
+  shell:
+    oc get all -n load-aware -oyaml > "{{ artifact_extra_logs_dir }}/all_default_resources.yaml"
```
kpouget, line 28: you can also save the `status`:
```
oc get all -n load-aware > "{{ artifact_extra_logs_dir }}/all_default_resources.status"
```
it's easier to understand at a glance, where reviewing the execution artifacts

and for the resource you plan to parse ==> store them in json, that's way faster to parse than Yaml, when there are many resources:
```
oc get all -n load-aware -ojson > "{{ artifact_extra_logs_dir }}/all_default_resources.json"
```
(and better store each type in a dedicated file)

drewrip:
The visualization parser hasn't been able to find the relevant artifacts that provide data for the default plots. This last commit should fix the last artifact that it couldn't find.

[parser log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1674780165319692288/artifacts/light/test/artifacts/001__plots/_matbench_parse.log)

drewrip:
[plots are generating](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1674800065614450688/artifacts/light/test/artifacts/001__plots/report_02_report:_control_plane_nodes_load.html)

[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1674800065614450688/artifacts/light/test/artifacts/000__load_aware_scale_test/001__loadaware__scale_test/_ansible.log).

@kpouget Should be good to merge!

testing/load-aware/config.yaml
```diff
@@ -96,6 +102,10 @@ load_aware:
   target_utilization: null
   safe_variance_margin: null
   safe_variance_sensitivity: null
+  scale_test:
+    distribution: null
+    duration: null
+    instances: null
```
kpouget, line 108: will the test work with `null` values?
if not (my guess, but I didn't check it all), please put default values that will run a small test (=make `small_poisson_load` the default :))

this way I can run the test without any preset to make sure that I didn't break anything

drewrip: I believe it should since I provided default parameters to the scale test function in `toolbox/load_aware.py` :
```
def scale_test(self, distribution="poisson", duration=60.0, instances=10):
```

I can add defaults to the config.yaml though if I'm wrong or if that'd be preferred :)

testing/load-aware/test.py
```diff
@@ -74,29 +74,29 @@ def _run_test(test_artifact_dir_p):
...
 
+        run.run("./run_toolbox.py cluster reset_prometheus_db")
+
         failed = True
         try:
-            run.run("./run_toolbox.py cluster reset_prometheus_db")
 
-            logging.info("Waiting 5 minutes to capture some metrics in Prometheus ...")
-            time.sleep(5 * 60)
+            run.run("./run_toolbox.py load_aware scale_test")
```
kpouget, line 91: hum, I knew something was wrong :) 
You need `from_config` to run you command the settings from `command_args.yaml`
```suggestion
            run.run("./run_toolbox.py from_config load_aware scale_test")
```
otherwise, only this ^^^ command is executed, as shown by [this file](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1674800065614450688/artifacts/light/test/artifacts/000__load_aware_scale_test/001__loadaware__scale_test/_python.cmd)

and 
```
  scale_test:
    distribution: null
    duration: null
    instances: null
```
this won't work, because it will pass the value `None` to the Python call, which isn't the same as `not passing a parameter` (only in this case the Python default value is used)

testing/load-aware/config.yaml
kpouget: see my other comment,
* currently these values _are_ used, because you're calling the command without any parameter
* but the config file settings aren't used either
* when you'll use `from_config` to pass the values from the config file ... your command will crash, because it will receive `None` values for the settings above

testing/load-aware/test.py
drewrip: Ah ok, I didn't realize how `from_config` worked entirely. I'll add the correct defaults to the config.

drewrip:
[tests passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1675815405597233152/artifacts/light/test/artifacts/000__load_aware_scale_test/001__loadaware__scale_test/_ansible.log)

[plots generated](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/810/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1675815405597233152/artifacts/light/test/artifacts/001__plots/report_02_report:_control_plane_nodes_load.html)

@kpouget Fixed the default configuration issue. Should be good to merge now. I also fixed the defaults for the `deploy_trimaran` role as well since it suffered from the same issue.

kpouget:
thanks @drewrip , I rebase and merge

PR #829, opened 2023-07-05
Roles: finish updating the roles location

kpouget:


kpouget:
Merging. Will be tested in the follow up PRs

PR #832, opened 2023-07-05
load-aware: scale-test runs coreutils make workload

drewrip:
The pod that runs the tests for coreutils is included, however it does not get included in the workload yet. If we want to have the tests run right after the build, it isn't straight forward but using pure pods. If we want to have some ratio of workloads be tests and some be builds, then we need to decide how to divy that up.


drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/832/pull-ci-openshift-psap-ci-artifacts-main-load-aware-e2e/1676719166565912576/artifacts/e2e/test/artifacts/000__load_aware_scale_test/001__load_aware__scale_test/_ansible.log). Should be good to merge.

[plots](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/832/pull-ci-openshift-psap-ci-artifacts-main-load-aware-e2e/1676719166565912576/artifacts/e2e/test/artifacts/001__plots/report_01_report:%20sutest%20cluster%20cpu-memory%20usage.html) showing spikes in CPU load.

roles/load_aware/load_aware_scale_test/files/load-aware-ns.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: load-aware
+  name: coreutils
   labels:
-    name: load-aware
+    name: coreutils
```
kpouget, line 6: I'm not sure why you're changing the namespace, I preferred the previous `load-aware` name 🤔 
a better solution would be to make it configurable in the config file, and here use the value from parameter

kpouget: > here

here and everywhere in the role :)

roles/load_aware/load_aware_scale_test/files/scheduler.py
```diff
@@ -1,44 +1,36 @@
...
-
-    yaml_pod_config = yaml.dump(pod_config, default_flow_style=False)
-    start_command = ["oc", "apply", "-n", "load-aware", "-f", "-"]
-    print(f"launching workload n{n} at {time.time()}")
-    p = sp.run(start_command, input=yaml_pod_config.encode(), stdout=sp.PIPE)
-    
+    make_pod = make_pod_config.format(n, scheduler_name)
+    start_command = ["oc", "apply", "-f", "-"]
+    print(f"launching make-pod n{n} at {time.time()}")
+    p = sp.run(start_command, input=make_pod.encode(), stdout=sp.PIPE)
```
kpouget, line 22: as a general programming rule, try to avoid single-char variables names, it's too easy to mix them up!
only exception might be `i`, although I often turn it into `idx`

roles/load_aware/load_aware_scale_test/files/load-aware-ns.yaml
drewrip: The `coreutils` namespace was originally created as a place to store the intermediate imagestreams and buildconfigs to be used in both of the tests. The `load-aware` namespace was used used for the workload pods. This caused issue since the workload pods don't have access by default to the imagestreams in the other namespace. I put them all in the same namespace, it just so happen to be the coreutils one.

I can make this configurable though.

roles/load_aware/load_aware_scale_test/files/scheduler.py
drewrip: Gotcha!

kpouget:
thanks for the updates @drewrip, looks good, I'll merge it if the test passes

one thing to consider when writting the Ansible roles: 
- what happens if the test fails, do you capture enough files to understand what happened without loosing too much time?
- if the test fails, can you run it again to give it another try?
here, this first task will fail if the namespace already exists (eg because it wasn't deleted at the end of the previous test ... because it failed)
```
- name: Create test namespace
  command:
    oc create ns {{ load_aware_scale_test_namespace }}
```
so either you start with
```
oc delete ns {{ load_aware_scale_test_namespace }} --ignore-not-found
```
or you create the namespace with:
```
oc create ns {{ load_aware_scale_test_namespace }} --dry-run=client -oyaml | oc apply -f-
```
(but no need to solve that now ^^)

kpouget:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/832/pull-ci-openshift-psap-ci-artifacts-main-load-aware-e2e/1677014973731049472/artifacts/e2e/test/artifacts/001__plots/reports_index.html), merging

PR #840, opened 2023-07-07
load-aware: build coreutils images in prepare step

drewrip:


kpouget:
[failure](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/840/pull-ci-openshift-psap-ci-artifacts-main-load-aware-e2e/1677282596230270976/artifacts/e2e/create-clusters/artifacts/sutest_ocp_install.log) is a flake :/

drewrip:
Tests passed but there was a silent failure. Needed to clean up pods between tests of the schedulers.

roles/load_aware/load_aware_scale_test/tasks/main.yml
```diff
@@ -1,77 +1,13 @@
...
-    oc get istag/coreutils-make:latest -n {{ load_aware_scale_test_namespace }}
-      | awk 'NR > 1 {print $1}'
-  register: make_image
-  delay: 15
-  retries: 80
-  until: make_image.stdout == "coreutils-make:latest"
-
 - name: Generate load timeline
   shell:
     python3 {{ load_timeline_generator }} {{ load_aware_scale_test_distribution }} {{ load_aware_scale_test_duration}} {{ load_aware_scale_test_instances }} > "{{ artifact_extra_logs_dir }}/load_aware_load_timeline.log"
```
kpouget, line 3: suggestion:
- take the destination file as a parameter
- write the timeline file in YAML (lists of dicts of dicts and `yaml.dump(timeline, f)`)

roles/load_aware/load_aware_scale_test/tasks/main.yml
```diff
@@ -1,77 +1,13 @@
...
-- name: Cleanup pods from default test
-  command:
-    oc delete pods --all -n {{ load_aware_scale_test_namespace }}
-
-- name: Run test pod with trimaran scheduler and load timeline
+- name: Run test workload with scheduler and load timeline
   shell:
     cat "{{ artifact_extra_logs_dir }}/load_aware_load_timeline.log"
-      | python3 {{ pod_start_scheduler }} trimaran {{ load_aware_scale_test_namespace }} > "{{ artifact_extra_logs_dir }}/trimaran_workload_scheduler.log"
+      | python3 {{ pod_start_scheduler }} {{ load_aware_scale_test_scheduler }} {{ load_aware_scale_test_namespace }} > "{{ artifact_extra_logs_dir }}/workload_scheduler.log"
```
kpouget, line 8: suggestion:
- take the timeline as a parameter, instead of stdin
- if stdout is really a log file, good to keep it like that

testing/load-aware/command_args.yaml
```diff
@@ -23,3 +23,16 @@ load_aware scale_test:
   duration: {{ load_aware.scale_test.duration }}
   instances: {{ load_aware.scale_test.instances }}
   namespace: {{ load_aware.scale_test.namespace }}
+  scheduler: {{ load_aware.scale_test.scheduler }}
+
+utils build_push_image/deps:
+  namespace: {{ load_aware.scale_test.namespace }}
+  image_local_name: coreutils-deps
+  tag: latest
```
kpouget, line 31: suggestion:
```suggestion
  image_local_name: coreutils
  tag: deps
```
(`latest` isn't recommenced anymore as container image tag)

testing/load-aware/command_args.yaml
```diff
@@ -23,3 +23,16 @@ load_aware scale_test:
...
+  dockerfile_path: testing/load-aware/images/Containerfile.deps
+
+utils build_push_image/make:
+  namespace: {{ load_aware.scale_test.namespace }}
+  image_local_name: coreutils-make
+  tag: latest
```
kpouget, line 37: suggestion:
```suggestion
  image_local_name: coreutils
  tag: make
```

testing/load-aware/config.yaml
```diff
@@ -102,17 +107,21 @@ clusters:
   cleanup_on_exit: false
 load_aware:
   log_level: 1
-  plugin: "TargetLoadPacking"
-  default_requests_cpu: "2000m"
+  plugin: TargetLoadPacking
+  default_requests_cpu: 2000m
   default_target_requests_multiplier: "2"
```
kpouget, line 112: I think this should be an int, no?

toolbox/load_aware.py
```diff
@@ -58,9 +58,9 @@ def undeploy_trimaran(self):
 
     @AnsibleRole("load_aware_scale_test")
     @AnsibleMappedParams
-    def scale_test(self, distribution="poisson", duration=60.0, instances=10, namespace="load-aware"):
+    def scale_test(self, distribution="poisson", duration=60.0, instances=10, namespace="load-aware", scheduler="default"):
```
kpouget, line 61: you'll have to add a short description for all the parameters in the `Args` below

toolbox/load_aware.py
```diff
@@ -72,4 +72,8 @@ def scale_test(self, distribution="poisson", duration=60.0, instances=10, namesp
             print(f"Can't run scale test using unknown distribution: {distribution}")
             sys.exit(1)
 
+        if scheduler not in ("default", "trimaran"):
+            print(f"Can't run scale test with unknown scheduler: {scheduler}")
+            sys.exit(1)
+
```
kpouget, line 78: is there a way to check if the value of `scheduler` is correct?
something like this?
```
oc get scheduler/{scheduler}
```
if yes, then this could be at the top of the Ansible task file and removed from here

otherwise, never mind, we won't test it with another secondary scheduler than Trimaran anytime soon :D

kpouget (commented):
thanks @drewrip, looks quite good
I put some suggestions in the comments, but I think you can wait before addressing them.
We can merge this PR, and beginning of next week you work on getting something interesting from the visualization.
Let me know when you're good for merging.

And would be nice to kick a run with many Pods, like something that would take 20min/30min to complete (x2 for to get the data for schedulers). We'll launch that on an `e2e` cluster.

testing/load-aware/command_args.yaml
drewrip: Ah ok, that makes sense. I'll make that swap!

roles/load_aware/load_aware_scale_test/tasks/main.yml
drewrip: Makes sense, I'll fix that the beginning of this week.

toolbox/load_aware.py
drewrip: I'll have to give that a shot. It didn't occur to me that the scheduler itself was a resource.

kpouget:
```
oc get schedulers -oyaml
apiVersion: v1
items:
- apiVersion: config.openshift.io/v1
  kind: Scheduler
  metadata:
    name: cluster
  spec:
    mastersSchedulable: false
    policy:
      name: ""
  status: {}
kind: List
metadata:
  resourceVersion: ""
```
we'll see if you get more than one `scheduler` :)

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/840/pull-ci-openshift-psap-ci-artifacts-main-load-aware-e2e/1678345728302256128/artifacts/e2e/test/artifacts/). 

@kpouget It should be good to merge as is, but I can add some of the suggestions to this PR in a bit if you want.

toolbox/load_aware.py
drewrip: Looks like I'm only seeing the one with Trimaran deployed. I don't think the specific scheduler gets listed as a `scheduler` resource.

testing/load-aware/config.yaml
drewrip: Still not sure why, but looks like its supposed to be a string according to the source: https://github.com/kubernetes-sigs/scheduler-plugins/blob/6ef6f63a0826b6045dd955eac599c902e396457f/apis/config/types.go#L102C28-L102C34

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/840/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1678733638277009408/artifacts/light/test/artifacts/). Should be good to merge now @kpouget

roles/load_aware/load_aware_scale_test/tasks/main.yml
```diff
@@ -1,51 +1,12 @@
...
 
-- name: Run test workload with default scheduler and load timeline
+- name: Run test workload with scheduler and load timeline
   shell:
-    cat "{{ artifact_extra_logs_dir }}/load_aware_load_timeline.log"
-      | python3 {{ pod_start_scheduler }} default {{ load_aware_scale_test_namespace }} > "{{ artifact_extra_logs_dir }}/default_workload_scheduler.log"
+    python3 {{ pod_start_scheduler }} {{ load_aware_scale_test_scheduler }} {{ load_aware_scale_test_namespace }} "{{ artifact_extra_logs_dir }}/schedule_plan.yaml" "{{ artifact_extra_logs_dir }}/schedule_execution.yaml"
 
-- name: Wait for default workloads to finish
+- name: Wait for workloads to finish
```
kpouget, line 9: (for the next PR) you need to have this in a `block`, and `Dump info about scheduler resources` in `always`

kpouget:
thanks @drewrip , I rebased and I'll merge

PR #851, opened 2023-07-19
load-aware: mix workloads in scale test

drewrip:


drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/851/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1682311315588124672/artifacts/light/test/artifacts/003__plots/report_03_report:_scheduler_performance.html).

The time to prepare the first pod of each type on each node is now much closer to the rest, thanks to the image preloading. We can also see the workload pods being scheduled on their own test nodes in the [scale test pod logs](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/851/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1682311315588124672/artifacts/light/test/artifacts/000__load_aware_scale_test_default_scheduler/001__load_aware__scale_test/all_pods.json).

Should be good to merge @kpouget !

testing/load-aware/config.yaml
```diff
@@ -89,23 +87,13 @@ clusters:
     compute:
       dedicated: true
       machineset:
-        name: rhods-pipelines
-        count: null
+        name: workload-pods
+        count: 3
```
kpouget, line 91: can you make this `1` in the default configuration, and override with a higher node count in specific tests

maybe also, force it to `1` in the light test (`clusters.sutest.compute.machineset.count: 1`) so that your light test never runs with more than 1 node

... or ignore the first comment, only update the `light` profile, to have `3` as a default for all the `load-aware-e2e` tests 🤔 
as you prefer !

testing/load-aware/command_args.yaml
```diff
@@ -36,3 +46,40 @@ utils build_push_image/make:
...
+cluster preload_image/make:
+  namespace: {{ load_aware.scale_test.namespace }}
+  name: make
+  image: {{ scale_test_imagestream }}:make
+  <<: *sutest_preload_image
+
+cluster preload_image/sleep:
+  namespace: {{ load_aware.scale_test.namespace }}
+  name: sleep
+  image: fedora:latest
```
kpouget, line 65: can you use `registry.access.redhat.com/ubi8/ubi` ? this image should be already part of the cluster

and, just for information, using an image `shortname` *address* is (now) discouraged,
Docker made it the usual to use eg `fedora:latest`, which turned into `docker.io/lib/fedora:latest`
but that's making you think that `docker.io` is the only image registry!

```
podman pull fedora:latest
Resolved "fedora" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.fedoraproject.org/fedora:latest...
Getting image source signatures
Copying blob ad5077952f52 [========>-----------------------------] 17.4MiB / 70.1MiB
```

so for `fedora`, the correct image value is: `registry.fedoraproject.org/fedora:latest`

kpouget: (and when you update to `ubi`, please also update the comment somewhere else in the PR)

visualizations/load-aware/plotting/pod_times.py
```diff
@@ -17,6 +17,8 @@ def register():
 def generatePodTimes(entry):
     data = []
     for p in entry.results.pods_info:
+        # We will need to generate different plots for each workload type in the future
+        if "sleep" in p.pod_name: continue
```
kpouget, line 21: 🤔 
not sure about  this `continue` 🤔 

one thing you can do is add a "configuration" flag, eg:
1. in the report:
```
                header += Plot_and_Text(f"...", set_config(dict(include_sleep=False), args))```
                header += Plot_and_Text(f"...", set_config(dict(include_sleep=True), args))```
```
2. in the plot:
```
cfg__instance = cfg.get("include_sleep", False)

...
if not include_sleep and "sleep" in p.pod_name: continue
```

the rational is that one way or another, I'd like to know about the scheduling of the sleep pods, as they will influence the scheduling of the build pods
"one way or another" because I'm not sure what's the best way to show them :D

testing/load-aware/config.yaml
drewrip: Can do, I'll probably update the light profile

testing/load-aware/command_args.yaml
drewrip: Makes sense, I'll switch over to `ubi`. Should I ensure it's part of the cluster by preloading it still?

> (and when you update to ubi, please also update the comment somewhere else in the PR)
Not entirely sure what you mean?

visualizations/load-aware/plotting/pod_times.py
drewrip: My original plan was create separate plots for each workload type. Each workload pod would get its own label and then plots for the time to complete for each workload would be added to the scheduler performance report. I'll probably try this to start.

I don't think a combined view of 2+ different workloads in one histogram would provide much information, but maybe it can show the overall performance across all workloads? Combining 2+ variables in one histogram would make it tough to draw conclusions I think

testing/load-aware/command_args.yaml
kpouget: > Should I ensure it's part of the cluster by preloading it still?

yes, for consistency, keep the preload code 


>>    (and when you update to ubi, please also update the comment somewhere else in the PR)

> Not entirely sure what you mean?

I'll add another comment to show you ;-)

testing/load-aware/test.py
```diff
@@ -62,15 +62,23 @@ def prepare_ci():
...
     run.run("./run_toolbox.py from_config cluster capture_environment --suffix sample")
     run.run("./run_toolbox.py from_config load_aware deploy_trimaran")
 
-    test_namespace = config.ci_artifacts.get_config("load_aware.scale_test.namespace")
-    run.run(f"oc create ns {test_namespace}")
     run.run("./run_toolbox.py from_config utils build_push_image --suffix deps")
     run.run("./run_toolbox.py from_config utils build_push_image --suffix make")
+    run.run("./run_toolbox.py from_config cluster preload_image --suffix deps")
+    run.run("./run_toolbox.py from_config cluster preload_image --suffix make")
+    run.run("./run_toolbox.py from_config cluster preload_image --suffix sleep") # The sleep image is just fedora:latest
```
kpouget, line 81: @drewrip , here is reference to `fedora:latest` that should be removed ;-)

visualizations/load-aware/plotting/pod_times.py
kpouget: > I don't think a combined view of 2+ different workloads in one histogram 

hum, yes, sorry, in the histogram plot it doesn't make any sense to show the `sleep` Pod 😆  🤦🏻 
```
data = generatePodTimeline(entry)
```
ok, the timeline data is generated from another function, so forget about it, it's all good !

testing/load-aware/test.py
drewrip: Gotcha!

drewrip:
@kpouget [test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/851/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1682411932352516096/build-log.txt). Plots are now generated for each type of workload in the report. The timeline combines all types of workloads.

This should include all of your suggestions!

kpouget:
thanks Drew, looks good :)  [Artifacts are there](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/851/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1682411932352516096/artifacts/light/test/artifacts/002__load_aware_scale_test_trimaran_scheduler/).

Merging :)

PR #873, opened 2023-07-27
cluster_deploy_operator: deploy_cr: add a namespace parameter when deploying the CR

kpouget:
…

otherwise it will be deployed in the active namespace

and that won't work in the CI


kpouget:
tested locally, merging.

PR #875, opened 2023-07-27
WIP: roles to deply and undeploy kepler

drewrip:


roles/cluster/cluster_deploy_kepler/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
+- name: Grab kepler-operator source
+  command:
+    git clone https://github.com/sustainable-computing-io/kepler-operator "{{ artifact_extra_logs_dir }}/kepler"
```
kpouget, line 3:
```suggestion
    git clone https://github.com/sustainable-computing-io/kepler-operator --quiet "/tmp/kepler"
```
`--quiet` will make the logs more readable
`"/tmp/kepler"` is to avoid cluttering the artifact dir with unnecessary files

maybe you can add this command:
```
git --git-dir "/tmp/kepler/.git" show
```
to track which commit was deployed

roles/cluster/cluster_deploy_kepler/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+  shell:
+    cd "{{ artifact_extra_logs_dir }}/kepler"; make deploy OPERATOR_IMG=quay.io/sustainable_computing_io/kepler-operator:latest
+
+- name: Deploy kepler exporters
+  shell:
+    oc apply -k "{{ artifact_extra_logs_dir }}/kepler/config/samples/"
+
+- name: Ensure kepler was deployed
+  shell:
+    oc get -n openshift-kepler-operator svc -oname
```
kpouget, line 15: why is it deploying in `openshift-kepler-operator` ? 🤔 
_this_ is the reason why I was seeing a different behavior, `openshift-` namespaces are trusted by default, they're not "user" namepsaces
(but I don't see which command tells it that it's an OpenShift cluster 🙃 )

roles/cluster/cluster_deploy_kepler/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+  register: kepler_deployed
+  delay: 5
+  retries: 30
+  until: kepler_deployed.stdout == "service/kepler-exporter-svc"
+
+- name: Create the service monitor to connect Kepler metrics
+  command:
+    oc apply -f {{ kepler_service_monitor }}
```
kpouget, line 23: you can move this one above `Ensure kepler was deployed`, so that Prometheus can discover the Kepler pod at the same time as your task!

roles/cluster/cluster_undeploy_kepler/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+- name: Grab kepler-operator source
+  command:
+    git clone https://github.com/sustainable-computing-io/kepler-operator "{{ artifact_extra_logs_dir }}/kepler"
```
kpouget, line 3: same as above for the destination directory

roles/cluster/cluster_undeploy_kepler/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+
+- name: Undeploy kepler-operator
+  shell:
+    cd "{{ artifact_extra_logs_dir }}/kepler"; make uninstall; make undeploy &2>1
+
+- name: Delete kepler-monitor ServiceMonitor
+  shell:
+    oc delete -n openshift-kepler-operator servicemonitor kepler-monitor --ignore-not-found
```
kpouget, line 11: you don't need `Delete kepler-monitor ServiceMonitor`, deleting the namespace will delete all of its content

but you _do_ need `Undeploy kepler-operator`, because Kepler may create Cluster-wide resources, which are not deleted when the namespace is deleted.

or `make uninstall; make undeploy` deletes the namespace it created? 🤔  not impossible (but it doesn't cost much to keep the `Delete kepler namespaces` task.

kpouget:
thanks @drewrip , looks good :) 
I made minor comments

testing/load-aware/test.py
```diff
@@ -74,6 +74,8 @@ def prepare_ci():
     run.run("./run_toolbox.py from_config cluster capture_environment --suffix sample")
     run.run("./run_toolbox.py from_config load_aware deploy_trimaran")
 
+    run.run("./run_toolbox.py from_config cluster deploy_kepler")
```
kpouget, line 77:
```suggestion
    run.run("./run_toolbox.py cluster deploy_kepler")
```
🙃

drewrip:
@kpouget  thanks for the comments! Since [the light test failed](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/875/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1684863053474041856/build-log.txt) and the upstream deployment via yaml files is causing its own issues, I'll try the OperatorHub installation again (which would be much simpler if it were to work). I'll come back and fix the manual version if it doesn't work though.

drewrip:
[test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/875/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1686357565690089472/artifacts/light/prepare/artifacts/006__cluster__deploy_kepler/_ansible.log). Kepler was deployed and its metric exporters were created.

The deploy and undeploy roles should include suggestions as given above in this thread.

The key issue was that the [kepler deployment makefile and yaml](https://github.com/sustainable-computing-io/kepler-operator/tree/v1alpha1) has a dependency on Go and hardcodes `kubectl`. The ci-artifacts image did not previously have either. I attempted resolving these inside the deploy role itself but ran into a number of issues. I resorted to installing them in the Containerfile itself. The best solution into the future would be to use the operator on OperatorHub.

jmencak:
> [test passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/875/pull-ci-openshift-psap-ci-artifacts-main-load-aware-light/1686357565690089472/artifacts/light/prepare/artifacts/006__cluster__deploy_kepler/_ansible.log). Kepler was deployed and its metric exporters were created.
> 
> The deploy and undeploy roles should include suggestions as given above in this thread.
> 
> The key issue was that the [kepler deployment makefile and yaml](https://github.com/sustainable-computing-io/kepler-operator/tree/v1alpha1) has a dependency on Go and hardcodes `kubectl`. The ci-artifacts image did not previously have either. I attempted resolving these inside the deploy role itself but ran into a number of issues. I resorted to installing them in the Containerfile itself. The best solution into the future would be to use the operator on OperatorHub.
> 
> @jmencak Whenever you get a second could you let me know what you think?

Would you be able to address all Kevin's comments?  Also, can we make "/tmp/kepler" into a variable so that it is not hardcoded like that at several places?

roles/cluster/cluster_deploy_kepler/tasks/main.yml
drewrip: I added showing the git commit that we are using, but using `--quiet` caused more complications than it's probably worth. Looks like `--quiet` puts the process in the background as well, surprisingly different than `wget`, so its more difficult to track when cloning is done.

roles/cluster/cluster_deploy_kepler/tasks/main.yml
drewrip: I made this move and occasionally this causes an error because the `openshift-kepler-operator` hasn't been deployed yet. I switched back to the way it was, but I could retry deploying it a couple times too.

roles/cluster/cluster_undeploy_kepler/tasks/main.yml
drewrip: Done!

roles/cluster/cluster_undeploy_kepler/tasks/main.yml
drewrip: I removed `Delete kepler-monitor ServiceMonitor` since this is definitely unneeded. The rest seems difficult to know for sure, especially since it is difficult to keep track of which resources specifically the makefile is deploying, and what resources might be added in the future.

kpouget:
Kepler got [deployed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift-psap_ci-artifacts/875/pull-ci-openshift-psap-ci-artifacts-main-load-aware-e2e/1688513927010324480/artifacts/e2e/prepare/artifacts/006__cluster__deploy_kepler/_ansible.log), but kepler logs didn't reach the Prometheus database. 

Merging as is, will fix it when we'll want to use Kepler.

kpouget:
for the record, here is the command I tried to deploy Kepler from OperatorHub (but it didn't work either):
```
./run_toolbox.py cluster deploy_operator  community-operators kepler-operator all
CSV=$(oc get csv -n openshift-operators -l operators.coreos.com/kepler-operator.openshift-operators -oname)
oc get $CSV '-ojsonpath={.metadata.annotations.alm-examples}' -n openshift-operators | jq .[0] | oc apply -f- -n default

curl -Ssf https://raw.githubusercontent.com/sustainable-computing-io/kepler-operator/v1alpha1/config/prometheus/monitor.yaml | yq | jq '.spec.selector.matchLabels = {"app.kubernetes.io/name": "kepler-exporter"} | .metadata.namespace = "default"' | oc apply -f-
oc label namespace default openshift.io/user-monitoring=true --overwrite
```

